PR #8, opened 2015-03-23
Update freeciv server to r28601

kvilhaugsvik:
I'm currenty unable to test this my self.

Test: found a city somewhere you are allowed to found a city. Don't get a message explaining why it isn't legal.


andreasrosdal:
I have tested this pull request, and verified that founding cities still works.

PR #36, opened 2016-07-25
Quick fix unit multi select problems.

kvilhaugsvik:
This fixes the right click to multi select problems I was able to find in Chromium Version 51.0.2704.79 Built on 8.4 running on Debian 8.5 (64-bit) and in Firefox 45.2.0 on Debian Stable.

I'm not familiar enough with Freeciv-web's UI code to know if this is the best way to fix the problem. This change may break platforms the current code works on. It is therefore sent as a pull request rather than directly committed to master.

### DIFF
--- a/freeciv-web/src/main/webapp/javascript/control.js
+++ b/freeciv-web/src/main/webapp/javascript/control.js
@@ -1354,6 +1354,10 @@ civclient_handle_key(keyboard_key, key_code, ctrl, alt, shift, the_event)
       /* Abort started multiple unit selection. */
       map_select_active = false;
       map_select_check = false;
+
+      /* Abort any context menu blocking. */
+      context_menu_active = true;
+      $("#canvas").contextMenu(true);
       break;
 
     case 32: // space, will clear selection and goto.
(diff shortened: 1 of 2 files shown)


andreasrosdal:
I think it is great that you have investigated this problem, and I have merged the pull request, but I added one [change](https://github.com/freeciv/freeciv-web/commit/44747e08da8d70fb876cb0cdea7710c19cbe06af).

The reason I did add the [change](https://github.com/freeciv/freeciv-web/commit/44747e08da8d70fb876cb0cdea7710c19cbe06af), was because as the pull request originally was, I would get a conext-menu popup when releasing the right mouse button after the drag-and-select multiple units action. I don't think there should be a context menu immediately after having selected multiple units. This is on Chrome and Firefox on Windows on. 

Are there any mouse control problems in the current master branch for you now?

(I coudn't find a convenient way to checkout the pull-request to my local system, so I just merged it, since the code looked quite good.)

kvilhaugsvik:
> the change,
No problem. (The change you reverted was supposed to support hypothetical custom context menu buttons. I have no idea if they exist or not. I have no idea if Freeciv-web tries to support them.)

> I don't think there should be a context menu immediately after having selected multiple units.
I agree. It didn't happen here. Could this be Windows specific jQuery behavior? (Since this is consistent across browsers I don't think the browsers are to blame)

> Are there any mouse control problems in the current master branch for you now?
Everything is working fine here. Tested both in Chromium and Firefox.

Thank you for reviewing and testing my change.

PR #43, opened 2016-11-12
WebGL: Use a better resolution and algorithm for the relief

lmoureaux:
spectacle m30211
This overhauls create_heightmap to use 16 vertices/tile.
The base height of a tile is the sum of two components: the distance to the sea and the terrain intrinsic height.
The tile height are then interpolated with a weight chosen to have smooth landscapes where the elevation doesn't vary much and steep slopes where it does.


andreasrosdal:
This looks awesome! Thanks!

PR #58, opened 2017-01-28
Cleanup servlet code

adaxi:
Github can ignore [whitespace](https://github.com/blog/967-github-secrets) when showing diffs. It might be useful when reviewing this pull request.

I used [Eclipse Copyright Tool](https://wiki.eclipse.org/Development_Resources/How_to_Use_Eclipse_Copyright_Tool) to fix the licences. It altered a bit the existing formatting, but on the other side it ensures that no file is forgotten.

Changes:
* Updated copyright dates
* Added license where license was missing.
* Removed useless @SuppressWarnings("unchecked")
* Replaced complex if condition with switch in CivclientLauncher
* Shortened long lines for easier readability on smaller screens
* Instead of building the query by concatenating strings now we use statement.setString
* Reply with 405 when using the GET method instead of 200.
* Renamed some variables to be compliant with Java conventions.
* Indented code. Please ignore whitespaces for more readable diff (-w)
* Removed duplicate code where possible. (see stats servlet)

### DIFF
--- a/freeciv-web/src/main/java/org/freeciv/servlet/CivclientLauncher.java
+++ b/freeciv-web/src/main/java/org/freeciv/servlet/CivclientLauncher.java
@@ -1,23 +1,20 @@
-/**********************************************************************
-    Freeciv-web - the web version of Freeciv. http://play.freeciv.org/
-    Copyright (C) 2009-2015  The Freeciv-web project
-
-    This program is free software: you can redistribute it and/or modify
-    it under the terms of the GNU Affero General Public License as published by
-    the Free Software Foundation, either version 3 of the License, or
-    (at your option) any later version.
-
-    This program is distributed in the hope that it will be useful,
-    but WITHOUT ANY WARRANTY; without even the implied warranty of
-    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-    GNU Affero General Public License for more details.
-
-    You should have received a copy of the GNU Affero General Public License
-    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
-***********************************************************************/
-
-
+/*******************************************************************************
+ * Freeciv-web - the web version of Freeciv. http://play.freeciv.org/
+ * Copyright (C) 2009-2017 The Freeciv-web project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *******************************************************************************/
 package org.freeciv.servlet;
 
 import java.io.*;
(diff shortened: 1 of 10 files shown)


ghost:
Thanks for the contribution to Freeciv-web! Feel free to make more!

adaxi:
Hi @andreasrosdal,

Civilization II is one of my all time favourite games. I discovered this project on HN and was interested at how it worked behind the scenes.

I found some vulnerabilities in the code base, I will try to send some pull requests to fix those. Before doing that I was wondering if I could get some info on the history of the project. 

I was wondering why are some game actions are handled by Java servlets, while the meta server is written in PHP code? I went a bit back in the repository history and they seem to coexist together for a long time. They also seem to share the same database. I wonder if it's a good idea to choose a single stack, instead of two. One one hand: it could reduce the number of dependencies to install the server, it could save some memory by having either Java or PHP running on the server, and we could perhaps reuse some code between both code bases. On the other hand, the current code structure clearly denotes a separation of concerns, and allows the freeciv-web community to learn both stacks.

ghost:
Pull requests to fix vulnerabilities are always welcome.

The metaserver is implemented in PHP and has origins back to the Freeciv metaserver. It handles communication between the Freeciv C servers and Freeciv-web database and metaserver. The source code can be found here: https://gna.org/projects/freecivmetaserve/ and the other production server can be found here:   http://meta.freeciv.org/

The Java servlets in Freeciv-web are implemented by me and are used by the Freeciv-web front-end.
I knew Java better than PHP, and I had to prioritize.

I agree that having a single stack would be very nice, and is something that I have thought about several times. However, in this Freeciv-web project I have tried to focus on making improvements which are directly requested by the users, and they have not requested this yet. If you are interested in helping to create a single stack for Freeciv-web, then that would be very welcome! Still, do you think that a single technology-stack is what would improve the game the most for end-users?

adaxi:
@andreasrosdal,

Thank you for the summary. 
I fully understand that in order for the project to succeed, sometimes sub-optimal choices are made due to lack of time or knowledge.
I also agree that the priority should be put on improving the game for the end-users, and having a single stack realistically will not change the end-user experience at this moment. I would like to improve the backend, since it is in this area that I feel more comfortable. This will perhaps free more time for you and other developers to work on the WebApp.

Adding websockets to the freeciv server would also be an interesting challenge, but a bit out of my comfort zone.

ghost:
I think Java would be the best candidate for the single technology stack. I will gladly review pull-requests for this.

What do you think are the most important things to improve in the game to make it better for end-users?

 Do you have Freeciv-web running on Vagrant?

adaxi:
I would also agree on keeping the Java stack.

To be honest I haven't played enough. I don't think major features are missing. Perhaps it needs improvements on the UX/UI. I'll play a game and write down potential improvements. 

Yes, I have Freeciv-web running on Vagrant.

ghost:
Thanks for the reply. Do you have more information about the vulnerabilities? It would be nice to have them fixed quickly.

adaxi:
https://github.com/freeciv/freeciv-web/blob/develop/freeciv-web/src/main/webapp/meta/metaserver.php#L184

I quickly noticed potential SQL injections.
Also phpinfo typically should be removed, because it gives the attackers knowledge on your machine, and it is easier for them to find an exploit that would work.

ghost:
Thanks for reporting this. Do you have a suggestion for a fix?

Is the SQL injection also present in the original metaserver.php file which is the source of this code?
http://svn.gna.org/viewcvs/freecivmetaserve/trunk/metaserver.php?view=markup

adaxi:
In this case I see that we are using some kind of wrapper around the PDO library. Which does not seem to support placeholders. A quick solution would be to wrap the variables around [PDO::quote](http://php.net/manual/en/pdo.quote.php)

adaxi:
After a quick look, it would be better, to add a second argument to [fcdb_exec](https://github.com/adaxi/freeciv-web/blob/develop/freeciv-web/src/main/webapp/meta/php_code/fcdb.php#L57), which takes the values of the placeholders:

```php

function fcdb_exec($stmt, $parameters = array()) {
  global $fcdb_sel, $fcdb_conn;
  set_error_handler("fcdb_error_handler");
  switch ($fcdb_sel) {
    case "PostgreSQL":
      $res = pg_Exec($fcdb_conn, $stmt);
      if (!$res) {
	build_fcdb_error("I cannot run a statement: '$stmt'.");
      }
      break;
    case "MySQL":
      $res = $fcdb_conn->prepare($stmt);
      $res->execute($parameters);
      if (!$res) {
	build_fcdb_error("I cannot run a statement: '$stmt'.");
      }
      break;
  }
  restore_error_handler();
  return ($res);
}
```

We would also need to rewrite the queries to use placeholders.

ghost:
Thanks!  Here I have added some code which was in the original metaserver.php 
https://github.com/freeciv/freeciv-web/commit/40ae7d2d046020887892bd6781c7bab757e2371e

I think that the $host and $port variables are reasonably safe now. Rewriting all the PHP queries will take some time, and that time should be spent on more useful things (not PHP, but Java!)...

Would you by any chance be interested in helping port the metaserver from PHP to Java? 
Java is much more secure and I think it would be more productive to spend time improving the Java stack.

adaxi:
Unfortunately there are [more](https://github.com/freeciv/freeciv-web/blob/40ae7d2d046020887892bd6781c7bab757e2371e/freeciv-web/src/main/webapp/meta/metaserver.php#L379), in this case the ```$host``` and ```$port``` are obtained from the $server_port variable.

I am currently working on the rewrite, I am testing/re-learning some stuff.

ghost:
@cazfi Perhaps this could be a problem in the original Metaserver as well?

PR #59, opened 2017-02-12
Replaces old PHP meta server with a Java Servlet based meta server.

adaxi:
I would be awesome this pull request could be tested before going to production.


ghost:
Thanks for creating this great improvement to Freeciv-web! It will improve the maintainability and security of the codebase. I will give it a good test before releasing to production. If you  want to make  more improvements, that would be very welcome!

adaxi:
Next step will be to split the metaserver servlet into 2 different servlets.

ghost:
The new Java-based metaserver is now live in production!

I notice that we lost the game-info page about a specific server, which was generated by this code in the old PHP server:   https://github.com/freeciv/freeciv-web/blob/test/freeciv-web/src/main/webapp/meta/metaserver.php#L372

adaxi:
Oops... I'll send a new pull request with this.

adaxi:
I also see some weird behavior in my browser where half of the resources (css, images, javascript) fail to load on the metaserver page. Though the url of the resources is correct.

adaxi:
Unless you did something, the problem seems to have disappeared automatically. I cannot reproduce it anymore.

By the way,  does you have some sort of irc/slack/gitter for freeciv-web? It would be easier to communicate in some cases.

ghost:
In general I think the deployment today went great!  
The game-info page is not very critical, so it is not urgent to fix it. We should get it fixed eventually, but it is not urgent. I saw it before deploying to production.

> Unless you did something, the problem seems to have disappeared automatically. I cannot reproduce it anymore.

I have also seen this problem sometimes in the past. I'm not sure about the cause, but it doesn't happen often, so I don't worry too much about it. I don't think it is related to the deployment today.

> By the way, does you have some sort of irc/slack/gitter for freeciv-web? It would be easier to communicate in some cases.

The quickest way to reach me is by e-mail:  andreas.rosdal (at) gmail. The Freeciv project also has a forum, mailinglist, #freeciv IRC channel on Freenode, comments here on Github, etc. but I usually prefer e-mail.

PR #60, opened 2017-02-13
Add the game details page

adaxi:
The details page should be back with this commit.

After coding this I have noticed that it would be a nice idea to use jsp includes to manager the header/footer in separate files and include them on each page, instead of copy pasting the headers and footers everywhere.


ghost:
Thanks, this looks very good!

ghost:
Released to production. More pull requests welcome! 👍

Issue #61, opened 2017-02-18
Frontpage face-lift


adaxi:
Facelift Demo
-----------------

In a previous pull request I stated that I would like to improve the front page and the metaserver pages. 
I ended up improving the design before doing structural work on the JSP logic.

https://adaxi.github.io/freeciv-demo/

This design has some qualities that facilitate the reworking of the JSP logic.

Motivation
-------------

Currently each page defines its own CSS and Javascript, which results in some duplicate code and hard to reuse code. To address this issue I would like to include the common code using JSP include directives.

The first step is to split CSS that is used on all pages from the CSS that is being used on specific pages. The demo page already incorporates that idea. This allows us to create the following template that incorporates styles, navigation and copyright:

```jsp
<html>
  <head>
	<%@include file="includes/head.jsp"%> <!-- Bootstrap/jQuery/FA -->
        <script></script> <!-- additional scripts for current page -->
	<link></link> <!-- additional scripts for current page -->
  <head>
  <body>
    <%@include file="includes/header.jsp"%> <!-- Navigation -->

    <!-- The page content goes here -->

    <%@include file="includes/footer.jsp"%> <!-- Copyright -->
  </body>
</html>
```
In this process I also attempted to reduce custom css code and reuse bootstrap components as much as possible, hence the face-lift.

I am 100% _not_ a designer but I think it turned quite OK. I have tested this under Firefox/Chrome in Debian, and should perform correctly on mobile devices. I have not tested on other platforms.

Next step
------------------

Should I continue my effort with this and start working on the JSP pages?

Contributing
---------------
Feel free to comment, or provide additional pull requests:
https://github.com/adaxi/freeciv-demo

ghost:
Wow, this looks very nice! Please do continue on these improvements. Thanks!

adaxi:
Thanks!
Next step will then be to integrate these ideas for the front page. After that I'll do the same for the metaserver.

ghost:
Don't forget mobile/tablets and the ads also.

adaxi:
How should I handle the new Chinese translations required for the frontpage?

ghost:
I think that the Chinese pages will have to stay on the old design.

PR #62, opened 2017-02-26
Frontpage & Game List facelift

adaxi:
Big one incoming!

This pull request attempts to incorporate ideas from issue #61.

- The metaserver servlet is not resposible to display the games list, a dedicated servlet is now used for that purpose.
- All the head/header/footer are now included in all the pages through JSP includes. 
- When possible update the servlet output to JSON instead of CSV.
- It was tested for mobiles and tablets.
- meta directory does not exist anymore.
- chinese translations where incorporated into the new design: parts where no translations were not available were hidden.

Important notes for deployment:
- The database is now created using flyway, instead of by importing the dump.
- nginx configuration was updated, the language selection is now done by the servlets. (take care to remove references to $lang.
- Please review/test the changes before going to production.


ghost:
Very nice! This will be a good improvement!


Why is the database initialized with Flyway? How can I release this to production then?

The tests fail on the Travis CI build: https://travis-ci.org/freeciv/freeciv-web

adaxi:
It does not even reach the tests in Travis :-(, it fails do build the WAR. I'll have a look at this during the week.

Regarding the database, it will not impact production since the schema was not altered.
I suppose that to put it in production is:

1. Update the nginx configuration so that it does not take into account the language settings.
2. Update the ROOT.war.

Regarding the database:
Instead of doing:
```sh
mysql -u ${mysql_user} -p${mysql_pass} freeciv_web < ${basedir}/freeciv-web/src/main/webapp/meta/private/metaserver.sql
sudo -u ubuntu mvn compile flyway:migrate
```

Now you only need to do:
```sh
sudo -u ubuntu mvn compile flyway:migrate
```

You can see the changes in the travis and vagrant build scripts.

adaxi:
I almost forgot, I also improved the format of /mailstatus to return JSON format instead of a mix of html and json. The pbem process will need to be restarted. I could not easily test pbem game list in my test environment, i ended up mocking the request to /mailstatus it with a static file.

ghost:
This is a great improvement, thanks a lot!

adaxi:
If there is any issue that you detect during testing, just report it here I will try to fix it.

ghost:
This has been released to production now. I like it very much. 

More pull-requests to Freeciv-web welcome! What do you think we could improve in the actual game to make it better for the average player?

adaxi:
From what I seen in the reviews user seem to be complaining about bugs:
https://chrome.google.com/webstore/detail/freeciv/ldhdjhmbapbeafmhdoobnlldhfopfcgh/reviews

I suppose next step will be to play a lot, and attempt to fix all encountered bugs.

- When starting a WebGL game, the textures take a long time to download, and you don't know if it is working. Perhaps a progress bar indicating the progress would be helpful.
- We need to fix this issue of resources not downloading: http://imgur.com/a/8usWL, I could not start a game just now.
- This weekend I showed the WebGL version to my brother, and we tried to play without creating an account, but it was not possible, we ended up creating an account.
- When on the WebGL 3D view, it is not clear that an unit is active, perhaps the ring around the unit should have some animation, or something similar.

adaxi:
Also it seems we have a bug in the play by email finished games: https://play.freeciv.org/game/list?v=play-by-email, they have all 122 and 16 wins respectively.

ghost:
Yes, fixing bugs such as these would improve the game a lot.

Issue #30, opened 2016-03-05
JavaScript warnings reported by JSHint


andreasrosdal:
Freeciv-web now uses JSHint to report JavaScript errors and warnings.
These errors can be found by running  'mvn install' in the freeciv-web directory.
http://jshint.com/docs/

(This is a good task for new developers who know JavaScript and are interested in contributing to Freeciv-web).

JSHint found problems with 9 files
/vagrant/freeciv-web/src/main/webapp/javascript/mapctrl.js
   32,17: 'e' is already defined.
   56,17: 'e' is already defined.
   141,17: 'e' is already defined.
   322,14: 'i' is already defined.
   341,15: 'ptype' is already defined.
   343,55: Missing semicolon.
   344,32: 'punit' used out of scope.
   345,54: 'punit' used out of scope.
   347,54: 'punit' used out of scope.
   353,9: 'punit' used out of scope.
   354,46: 'punit' used out of scope.

/vagrant/freeciv-web/src/main/webapp/javascript/savegame.js
   54,19: Implied eval. Consider passing a function instead of a string.
   203,41: Missing semicolon.
   286,99: Missing semicolon.
   410,14: 'i' is already defined.
   554,6: Missing semicolon.
   564,3: Implied eval. Consider passing a function instead of a string.

/vagrant/freeciv-web/src/main/webapp/javascript/packhand.js
   1157,17: 'pplayer' is already defined.

/vagrant/freeciv-web/src/main/webapp/javascript/pbem.js
   286,54: Missing semicolon.
   305,29: Missing semicolon.
   404,9: Implied eval. Consider passing a function instead of a string.
   431,30: Missing semicolon.
   448,5: Implied eval. Consider passing a function instead of a string.
   492,3: Implied eval. Consider passing a function instead of a string.
   505,18: Implied eval. Consider passing a function instead of a string.

/vagrant/freeciv-web/src/main/webapp/javascript/tech.js
   181,20: 'tech_id' is already defined.
   182,15: 'ptech' is already defined.
   196,53: Missing semicolon.
   213,15: 'tag' is already defined.
   219,53: Missing semicolon.
   227,19: 'bgcolor' is already defined.
   229,15: 'tag' is already defined.
   235,53: Missing semicolon.
   244,16: 'i' is already defined.
   250,16: 'i' is already defined.
   380,14: 'i' is already defined.
   438,17: 'e' is already defined.

/vagrant/freeciv-web/src/main/webapp/javascript/tilespec.js
[...]

PR #64, opened 2017-03-28
JSHint cleanups

lonemadmax:
WIP for https://github.com/freeciv/freeciv-web/issues/30 (just starting to get the hang of this, don't hold your breath).
Removes (most) JSHint warnings for 2dcanvas/mapctrl.js.
Uncovers and solves a real bug in units window that showed some info from the last unit in the stack, instead of the active one.

To clean the other two warnings, I'd have to rename the variables for their respective blocks or move their declarations to the outer function block. Or use [`let`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) instead of `var`, which I think may be The Right Way (block scope instead of function scope), but that's ES6, JSHint warns with `'let' is available in ES6 (use esnext option) or Mozilla JS extensions (use moz).`, and I don't know what freeciv-web targets and how to change it, if at all. Also, I found out about let when preparing this change (did I say I'm a novice and whatever I contribute must be thoroughly reviewed?)

### DIFF
--- a/freeciv-web/src/main/webapp/javascript/2dcanvas/mapctrl.js
+++ b/freeciv-web/src/main/webapp/javascript/2dcanvas/mapctrl.js
@@ -55,7 +55,7 @@ function mapview_mouse_click(e)
   var rightclick = false;
   var middleclick = false;
 
-  if (!e) var e = window.event;
+  if (!e) e = window.event;
   if (e.which) {
     rightclick = (e.which == 3);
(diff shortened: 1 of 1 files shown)


ghost:
Thanks a lot for the contribution to Freeciv-web. Feel free to create more improvements if you want. 

I have released this fix to the production server now.

ghost:
I don't think we should use let at this time.

PR #70, opened 2017-05-26
include bold, italic variants of Open Sans font

drdogbot7:
Currently bold text isn't well differentiated on the website because only the regular (400) variant of "Open Sans" is included. This will include the bold (700), italic (400i) and bold italic (700i) variants.

Example of bold text, almost indistinguishable:
image


ghost:
Thanks for the contribution to Freeciv-web.

PR #73, opened 2017-05-28
Vagrant provisioning enhancement for new boxes

lonemadmax:
Add a workaround for a bug in virtualbox guest additions 5.1.20 and a fix to provision new boxes after having used another one (both triggered after destroying previous yakkety and downloading new zesty).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -52,7 +52,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     v.customize ["modifyvm", :id, "--memory", mem]
   end
 
-  config.vm.synced_folder "./", "/vagrant"
+  # workaround vbox bug 16670
+  config.vm.synced_folder "./", "/vagrant", disabled: true
+  config.vm.provision :shell, :path => "scripts/vbox-workaround-16670.sh", run: "always"
 
   # run the Freeciv bootstrap script on startup
   config.vm.provision :shell, :path => "scripts/vagrant-build.sh", run: "always"
--- a/scripts/vagrant-build.sh
+++ b/scripts/vagrant-build.sh
@@ -28,8 +28,9 @@ echo "================================="
 echo "Running Freeciv-web setup script."
 echo "================================="
 
-# if Freeciv-web already built with Vagrant, then start it instead.
-if [ -f "/vagrant/freeciv-web/target/freeciv-web.war" ]; then
+# if system already provisioned and Freeciv-web already built with Vagrant,
+# then start it instead.
+if [ -f "/usr/sbin/nginx" -a -f "/vagrant/freeciv-web/target/freeciv-web.war" ]; then
   printf "\n\nFreeciv-web already built, starting it.\n\n-----";
   cd ${basedir}/scripts/ && sudo -H -u ubuntu ./start-freeciv-web.sh
   printf "Freeciv-web started. Now login with 'vagrant ssh' and point your browser to http://localhost";
--- a/scripts/vbox-workaround-16670.sh
+++ b/scripts/vbox-workaround-16670.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# Detect and correct bug in vbox guest additions 5.1.20
+# See https://www.virtualbox.org/ticket/16670
+
+[ -e "/sbin/mount.vboxsf" ] || sudo ln -sf /opt/VBoxGuestAdditions-5.1.20/lib/VBoxGuestAdditions/mount.vboxsf /sbin/mount.vboxsf
+mount -t vboxsf -o uid=`id -u ubuntu`,gid=`id -g ubuntu` vagrant /vagrant


ghost:
Thanks for your contribution to Freeciv-web!

ghost:
This bug has been fixed in VirtualBox 5.1.22.  Coudn't you just use that version?

lonemadmax:
It is already fixed in 5.1.22, but the image uses 5.1.20.

ghost:
Ok, thanks a lot for the fix. I hope you will contribute even more to this project.

ghost:
I tried a new 'vagrant up' from a fresh git checkout, and got this error message now:

$ vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'ubuntu-zesty64'...
==> default: Matching MAC address for NAT networking...
==> default: Setting the name of the VM: freeciv-web_default_1496084405933_48401
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
==> default: You are trying to forward to privileged ports (ports <= 1024). Most
==> default: operating systems restrict this to only privileged process (typically
==> default: processes running as an administrative user). This is a warning in case
==> default: the port forwarding doesn't work. If any problems occur, please try a
==> default: port higher than 1024.
==> default: Forwarding ports...
    default: 80 (guest) => 80 (host) (adapter 1)
    default: 443 (guest) => 443 (host) (adapter 1)
    default: 22 (guest) => 2222 (host) (adapter 1)
==> default: Running 'pre-boot' VM customizations...
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: ubuntu
    default: SSH auth method: password
    default: Warning: Remote connection disconnect. Retrying...
    default: Warning: Authentication failure. Retrying...
    default:
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if it's present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
==> default: Running provisioner: shell...
    default: Running: C:/Users/andre/AppData/Local/Temp/vagrant-shell20170529-3508-1a58hco.sh
**==> default: /sbin/mount.vboxsf: mounting failed with the error
==> default: : No such file or directory**
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.


This is running Windows 10, VirtualBox 5.1.22.

Perhaps you could help to resolve the issue?

ghost:
I also get this error message:



==> default: running install_egg_info
==> default: Writing /usr/local/lib/python3.5/dist-packages/mysql_connector_python-2.1.3.egg-info
==> default: ==== Setting up MySQL ====
==> default: mysqladmin:
==> default: [Warning] Using a password on the command line interface can be insecure.
==> default: /tmp/vagrant-shell: line 93: cd: /home/ubuntu/../freeciv-web: No such file or directory
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.

lonemadmax:
I'm at a loss.  Most I have found is https://github.com/mitchellh/vagrant/issues/8352, but I'm not sure it's exactly the same, and it's also solved in 5.1.20 (and a few intermediate others, it seems).

The second one would be a consequence of the first if they are on the same run or for some other reason the /vagrant directory was not mounted.  At the beginning of vagrant-build.sh (lines 10-19) there's a choice of basedir dependent on the existence of /vagrant, with the purpose, it seems, to make it work both when installing in an image or in a real machine.  When /vagrant does exist, line 93 that is giving the error would go to /vagrant/freeciv-web.  When it doesn't, it'd go to $(pwd)/../freeciv-web, the home directory of an inexistent freeciv-web, it seems.

I see you reverted the patches.  Did that solve it and the machine was provisioned?  In that case, I'm very sorry I broke your machine when making mine work.

The workaround is only needed once (per machine) and only when 5.1.20 executes. It shouldn't give problems when it's not needed, though, but if reverting makes your work, there's no question it should be done.

The change in vagrant-build.sh should be there (in one form or another) for the case of a new machine having used another one before.  How did your clean box get nginx & co. without it? Or did you delete freeciv-web/target/freeciv-web.war before bringing up the box?

ghost:
Thanks a lot for helping me with this issue. I think this is some kind of Vagrant problem, but I'm not sure about the root cause. I know your intentions were good.

Do you run Windows or Linux? I'm running Windows 10.

> I see you reverted the patches. Did that solve it and the machine was provisioned?

Yes, reverting it solved it on my system. I have tested 'vagrant up' on a clean checkout of Freeciv-web from Github, and that worked.

> How did your clean box get nginx & co. without it? Or did you delete freeciv-web/target/freeciv-web.war before bringing up the box?

I'm not sure. I will try again, and try to reproduce the problem.

lonemadmax:
> Do you run Windows or Linux? I'm running Windows 10.

Fedora Linux here.

Reading again vagrant-build.sh, it's clear how you got the packages: installation comes before the error, and it's independent of the existence of /vagrant folder.  So you updated the system (line 64), got the dependencies (l 69), installed tornado (l 78), wikipedia (l 80) and mysql-connector-python (l 87).  Then it failed just before running flyway:migrate.  You wouldn't need the last steps of building and testing freeciv-web, as you'll do it manually anyway when changing code, but I don't know about flyway:migrate and the other steps with configuration files.

> I know your intentions were good.

The intention was the typical "I've had this problem. Others may stumble into the same one. The solution is simple once you know it and looks riskless, so let's share it". I don't know how many people use the VM, but with the workaround only being needed when running a very specific version the first time, maybe it was not such a good idea to send a patch.

Do I guess right if I say you got the mount error, logged into that running box and manually run vagrant-build.sh?  In that case, I'd keep the test for nginx (or whatever other method to check whether the necessary packages are installed), as the reason for the second error would be the first one that didn't mount /vagrant.

PR #75, opened 2017-05-28
Add note about port 8080 on macOS/Linux

jacobrask:
Port 80 is forwarded to port 8080 on non-Windows systems, and it is fine to use that port for development.
The few people who want to run a production server will know how to set up an SSH tunnel.

I ended up struggling for a while with the SSH tunnel before realising I did not need it.


ghost:
Thanks!

Issue #69, opened 2017-05-16
Chat messages can't contain special characters like ' " <


ghost:
Chat messages can't contain special characters like ' " <. This was added to protect against XSS security issues. However, some users report this as a problem.

Some relevant code: 
https://github.com/freeciv/freeciv-web/blob/develop/freeciv-web/src/main/webapp/javascript/control.js#L358
One of the patches against the Freeciv server also disallows " and <>

PR #79, opened 2017-06-01
Bring apostrophes and other characters back to the message box

lonemadmax:
Allows sending apostrophes in the text. The problem was different in two places:
- the chatbox would remove the first one, changing "it's" to "its", and the server would cut the message if there was a second one
- the new private message dialog didn't even remove them, so the server cut the message when it found one.

This patch makes both ways use the same escaping, and replaces the "bad" characters by their html entities. As the message is going to be injected as html, that should be enough.

Solves issue #69.

### DIFF
--- a/freeciv-web/src/main/webapp/javascript/control.js
+++ b/freeciv-web/src/main/webapp/javascript/control.js
@@ -354,15 +354,27 @@ function update_mouse_cursor()
 }
 
 /****************************************************************************
+ Common replacements and encoding for messages.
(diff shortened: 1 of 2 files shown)


lonemadmax:
If ", <, > or & shouldn't be escaped because they may be used in messages to the server or something, feel free to change the corresponding lines.  I may not see any messages until the weekend.

ghost:
Thanks a lot for your contribution to Freeciv-web!

PR #85, opened 2017-06-19
Keep order in cities tab and cities navigation

lonemadmax:
These changes:
* Add the city id to the rows in the cities tab.
* Updates the cities tab when there are changes, instead of only when changing to that tab.
* Keep the cities tab order between updates.
* Use that order for the next/prev buttons in the city dialog.

Things to take into account:
* We just have the columns and direction by which the tabled is sorted, so when using several columns, this may not work. For example, sorting by production and granary will give you a table sorted by granary and production instead in the next update.
* Navigation from the city dialog calculates next and previous cities in the moment next/prev is clicked, not when the dialog was first open. That gives us new cities, removes lost cities and has all the other updates. So, if you wanted to sort by production to change all the cities building tanks to build currency, changing production in the first one will change its position in the table and when pressing next you won't get what you expected for this use case.

### DIFF
--- a/freeciv-web/src/main/webapp/css/civclient.css
+++ b/freeciv-web/src/main/webapp/css/civclient.css
@@ -918,7 +918,7 @@ button.fg-button {  width:auto; overflow:visible; } /* removes extra button widt
   width: 100%;
 }
 
-#cities_row {
+.cities_row {
   cursor:pointer; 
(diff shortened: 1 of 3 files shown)


ghost:
Thanks a lot for this great improvement to Freeciv-web. More contributions are welcome!

PR #86, opened 2017-07-06
Chatbox enhancements

lonemadmax:
* Keep focus in chat line after sending message.
* Better styles, obsoleting resize callback.
* A button to switch between sending messages to everybody and to allies.
* Change colors of some messages to improve readability.

### DIFF
--- a/freeciv-web/src/main/webapp/css/civclient.css
+++ b/freeciv-web/src/main/webapp/css/civclient.css
@@ -987,3 +987,14 @@ li.tech-both::before {
   left: 1em;
 }
 
+.bw-toggle[data-toggle=true] {
+  color: lightgrey;
+}
+
+.bw-toggle[data-toggle=false] {
+  color: grey;
+}
+
+.bw-toggle:hover {
+  color: white;
+}
--- a/freeciv-web/src/main/webapp/css/civclient.css
+++ b/freeciv-web/src/main/webapp/css/civclient.css
@@ -37,26 +37,28 @@ button {
 
 
 #game_message_area {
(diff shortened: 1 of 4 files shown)


ghost:
Thanks a lot. Awesome!

ghost:
This has been released to production now.

ghost:
There has been some reports that the allied chat is not working correctly in production. Perhaps you could investigate?

lonemadmax:
A special case for longturn. Pushed https://github.com/freeciv/freeciv-web/commit/7b618e5915c472d994262bbeae1f630e88449745. What is that special case for, anyway?

ghost:
Thanks for fixing this so quickly.

The special case for longturn in send_message() is for when messages in LongTurn games are shown from the event cache, for example a chat message in a LongTurn game. In that case, we want to show the player-name, since otherwise the playername will be missing from chat messages from the event cache. This is quite a poor fix, but it worked.

lonemadmax:
I see.  It's not some strange longturn thing, just it's not needed in normal games.

That hack is to go, then. I'm thinking of changing the handling of messages (it will take time, don't hold your breath) to be interpreted in the client instead of just injecting them as html, and making each of them a separate item instead of joining them as they are now, to allow filtering.

The names are in the event cache, but as &lt;player&gt;, which makes them disappear when injected as html.  It also happens during gameplay, you can see it in the inspector.

Mmmm...  Off to register a user with the name "script".

ghost:
Yes, the hack is to go. Thanks for improving this!!

PR #87, opened 2017-07-08
Show correct FPT in city dialog minimap.

lonemadmax:
When the server sends food_output & co, only valid tiles are iterated (city_tile_iterate). That means near the poles we don't get the 21 tiles we expect for radius_sq = 5, but only 18 or 13. That makes the city dialog show incorrect info for some workers, and not draw others.

This patch makes the client aware of that. The Right Way to solve it may be to change the server code, though, to always send info for 21 tiles, with 0 output for invalid ones. Or better yet, make the client calculate it, as other clients do, and get rid of misc_devversion_sync.patch.

before:
before

after:
after


ghost:
Thanks a lot! Keep these good improvements coming!

PR #132, opened 2017-11-13
disable civil war in PBEM

drdogbot7:
Setting civilwarsize to 1000 to disable civil war and avoid creating AI players


ghost:
Thanks

PR #141, opened 2017-12-23
Dockerfile Update

chris1869:
Dear freeciv enthusiasts,
I wanted to create a fully self-standing docker image for freeciv that can be easily installed. The below Dockerfile, incl. the respective changes to run scripts, should create such a docker image.

Further code might be required to be pushed into the Dockerfile, vs. changing actual scripts to ensure compatibility with current freeciv - distro.

Kindly let me know, if this functionality is of interest for you, so we can further discuss how to best approach compatibility issues. 

Best regards
Chris


ghost:
Thanks for creating this pull-request. Adding Docker support would be a great improvement.
I will add some specific comments about things which needs fixing: this is mostly related to
that the existing Vagrant support needs to be tested and verified that it still works.
Also, settings files should not be committed, only the dist files.

freeciv-proxy/settings.ini
```diff
@@ -0,0 +1,4 @@
+[Config]
+mysql_user = root
+mysql_database = freeciv_web
+mysql_password = freeciv_docker
```
ghost: This file should not be added. Only add settings.dist files, then copy the .dist file to settings.ini

pbem/settings.ini
```diff
@@ -0,0 +1,13 @@
+[Config]
```
ghost: This file should not be added. Use a .dist file instead.

publite2/settings.ini
```diff
@@ -0,0 +1,15 @@
+[Config]
+#Freeciv savegames are stored in this directory
```
ghost: Should not be added.

freeciv-web/src/main/webapp/META-INF/context.xml.dist
```diff
@@ -9,7 +9,7 @@
     (inside quatation marks) and rebuild. -->
   <Resource name="jdbc/freeciv_mysql" auth="Container" type="javax.sql.DataSource"
                maxTotal="100" maxIdle="30" maxWaitMillis="10000"
-               username="root" password="vagrant" driverClassName="com.mysql.cj.jdbc.Driver"
+               username="root" password="freeciv_docker" driverClassName="com.mysql.cj.jdbc.Driver"
```
ghost: This will break Vagrant support.

chris1869:
Thank you for the comments and suggestions. I'll remove the settings.ini files and script copying them from *.dist in the Dockerfile. Let me know, if you think, this should also be excluded and admins be required to copy *.dist files manually. I'll set back password to vagrant. In a later version mysql_pass should be an installation parameter, as it is used quite frequently throughout the installation scripts. Best Chris

freeciv-proxy/civcom.py
```diff
@@ -142,6 +142,7 @@ def send_buffer_append(self, data):
                 data.decode(
                     encoding="utf-8",
                     errors="ignore"))
+            logger.info("send to client: " + self.send_buffer[-1])
         except UnicodeDecodeError:
```
ghost: Can you please  change so that this logging of messages are not logged by default?

scripts/dependency-services-default-start.sh
```diff
@@ -8,7 +8,7 @@
 
 
 export JAVA_OPTS="-Djava.security.egd=file:/dev/urandom"
-export CATALINA_HOME=/var/lib/tomcat8
+export CATALINA_HOME=/usr/local/tomcat
```
ghost: This change will break Freeciv-web in production and on Vagrant.

freeciv-web/src/main/webapp/WEB-INF/config.properties
```diff
@@ -0,0 +1,5 @@
+captcha_secret=secret goes here
+savegame_dir=/var/lib/tomcat8/webapps/data/savegames/
```
ghost, line 2: This file should not be added. Use a .dist file instead.

freeciv-web/build.sh
```diff
@@ -24,4 +24,5 @@ mkdir -p $DATADIR/scorelogs
 mkdir -p $DATADIR/ranklogs
 
 echo "maven package"
+sudo mvn compile flyway:migrate
 mvn package && cp target/freeciv-web.war "${TOMCATDIR}/webapps/ROOT.war"
```
ghost, line 28: I'm not sure that this change should be added, since it will slow down building.

scripts/docker-build.sh
```diff
@@ -39,7 +41,6 @@ cd ${basedir}
 echo "==== Installing Updates and Dependencies ===="
 export DEBIAN_FRONTEND=noninteractive
 echo "mysql setup..."
-mysql_install_db
```
ghost: Why?

ghost:
Please test the Vagrant build and make sure it works.

freeciv-web/build.sh
chris1869: I know, it will slow down building. I introduced this step, for the DB migration steps to kick-in. Otherwise the relevant mysql tables will not be available and all servlets (e.g., meta server), will not work. At least to my understanding of how those components interact.

scripts/docker-build.sh
chris1869: Shifted to new script scripts/docker-prep_serv.sh with all other startup mysql code

ghost:
Thanks!

ghost:
@chris1869 I reverted this pull-request, since it broke the existing Vagrant support.
Please make sure that this change doesn't break the Vagrant support, then please submit the pull-request.

For example, your change in publite2/init-freeciv-web.sh
Requiring the path /docker/freeciv/freeciv/server/freeciv-web to exist will not work on Vagrant, only on Docker.

I really want Docker support to work, so please submit an update!

chris1869:
@andreasrosdal I just checked that there are no more references to /docker/* paths in any non-Docker related scripts. Update will follow. 

Is there a way to check Vagrant is not broken more thoroughly? Before the last update, I successfully ran Vagrant up and started playing 2-3 turns. Travis CI seemed also to run successfully. I don't want to break current code.

ghost:
Thanks, looking forward to the update.

> Is there a way to check Vagrant is not broken more thoroughly? 

Just running 'vagrant up' and starting a game on a fresh Freeciv-web directory should be enough to verify that Vagrant is working.

> Before the last update, I successfully ran Vagrant up and started playing 2-3 turns. Travis CI seemed also to run successfully. I don't want to break current code.

For example, the following change to publite2/init-freeciv-web.sh was included in the pull-request:

https://github.com/freeciv/freeciv-web/pull/141/files#diff-35104b44e320d93b4c6ccd0726b5f5fa

+/docker/freeciv/freeciv/server/freeciv-web  will never work on Vagrant.

PR #157, opened 2018-02-12
Add "Build farmland" option to irrigate a second time

Arctik314:
Having discovered "Refrigeration" tech, one should be able to build farmlands on irrigated tiles, which is translate by the game as a second irrigation.

Currently, the only way to perform this action is by pressing the "I" shortcut for irrigation.

This pull-request adds a "Build Farmland" option in the options menu of a unit that can irrigate, and a "Build Farmland" button, which uses the default button for irrigation.


ghost:
Thanks for your contribution to Freeciv-web!

PR #160, opened 2018-03-19
Change database to MSSQL and create DataSource in context from env var

zekoz:
Changes database to MSSQL and creates a resource factory for creating context DataSource object from environment variable connection string. This is in preparation for migration to azure app service.


ghost:
Thanks for this contribution to Freeciv-web. However, I don't agree with this change at this time.
Firstly, the Travis CI tests fail, so this has to be fixed.
Moreover, I don't think MSSQL should replace MySQL. MySQL is open source.
Perhaps support for MSSQL could be added as an optional database engine,
in addition to the current MySQL. MySQL should still be the default in this repository.
You could maintain a fork of Freeciv-web with MSSQL and other changes you need in your own repository.

PR #161, opened 2018-03-20
Adds cache filters to web.xml: 30 days expiry for images to reduce bandwidth cost.

zekoz:
This may be premature optimization, but the game assets are pretty large.


ghost:
I don't think all urls can be cached for 30 days.
Can you please check this?

PR #163, opened 2018-04-28
Unified install script

lonemadmax:
Merge the different install scripts into one, in the hope of:
- reducing the differences
- having just one place to change when updating dependencies or enhancing configuration
- making it easier to support installation in other systems


kvilhaugsvik (commented):
If I recall correctly systemctl asks for my password by default if I don't run it as root. Are there configurations where sudo is needed here?  (I'm asking. I haven't looked my self.)

kvilhaugsvik:
(The above question was meant for 650bff64cfdf6f31dc63c07941d4c3569f70c501)

PS! Unifying the install scripts is a great idea.

lonemadmax:
All I've tested with systemctl is a minimum Debian 9 install. It seems to be related to polkit policies, don't know the specifics, will try to find out something tomorrow.

I think the less sudo uses we have, the better. Even if the default in some systems is to require root, we might drop that part and force the user to either add a policy or tweak the scripts. After all, no one has opened a bug about that yet...

lonemadmax:
Systemd uses polkit and may ask for credentials through it when needed. Installing the policykit-1 package allowed me to run systemctl directly. It asked me for my password on every invocation, unlike sudo, but from what I've understood, that's something a sysadmin can change.

So I'm OK with reverting https://github.com/freeciv/freeciv-web/commit/650bff64cfdf6f31dc63c07941d4c3569f70c501 and removing the uses in  [3a64fa3a](https://github.com/freeciv/freeciv-web/pull/163/commits/3a64fa3a3f9ee3a2ba4c83a3313842dfa73e644f#diff-14404e0910d46c9b9e79591bd4602490R228).

Regarding policykit-1, I see three options:
1. Do nothing, it's the sysadmin responsibility. We don't install sudo, systemd, upstart, sysvinit, why should policykit-1 be different?
2. Add it to the dependency packages list. That would install it also when not needed. It would even slurp systemd if it was not in the system.
3. Add it to the list, but only when systemctl is in the system (maybe when systemctl is but service is not).

kvilhaugsvik:
Not sure what alternative is best. 1 could be combined with a warning. But that may on the other hand be as much work as 3.

lonemadmax:
Work is not an issue (at least for 3): `if which systemctl; then dependencies="${dependencies} policykit-1"`. The question is whether it would be rude to install policykit. On the other hand, we already overwrite nginx and tomcat config, so maybe I'm worrying for nothing.

There's another alternative: keep it in the install script, so it will work no matter what, and remove it from start/stop, so it will fail for the systemd-but-no-polkit case and force the admin to make a decission.

Wait, there's still another one: detect polkit (`which pkexec` should do) either in the install script (and point to different start/stop scripts) or in the start/stop ones (and wrap with sudo if needed):

**STILL NOT TESTED**
```bash
if which pkexec > /dev/null; then
  CREDENTIALS_MANAGER=
else
  CREDENTIALS_MANAGER=sudo
fi

for unit in mysql nginx tomcat8; do
  systemctl is-active --quiet ${unit}.service || ${CREDENTIALS_MANAGER} systemctl start ${unit}.service
done
```

PR #164, opened 2018-06-07
Introduce hack mode for web client only testing.

kvilhaugsvik:
In addition to code for playing Freeciv in the web browser Freeciv-web
includes infastructure and code to administrate and run a big installation.
Stuff like the meta server and Freeciv servers dedicated to certain game
types are examples of this.

The big installation stuff isn't always needed. This applies both to
extremely small installations and to developers testing Freeciv-web changes
that don't touch the big installation infrastructure. An example of the
former are a single game with a few players that happens to be power users.
An example of the latter is a developer testing a change that only touches
the web client and/or the Freeciv server.

When I test changes to Freeciv-web I often log in using the same user name
before I load a test save game from a limited set or a ruleset. Doing this
manually often takes more time than testing the change it self. Being able to
jump straight into a test game would, at least for me, be extremely useful.

The Freeciv-web meta server is, at least in my set up, fragile. It has often
broken without leaving any clues that have enabled me to find out what is
wrong. Being able to bypass the Freeciv-web meta-server would allow me to
test changes to, and therefore work on, Freeciv-web in times when it is down.

Introduce a "hack" value to the "action" URL parameter. It takes the port of
the Freeciv server from the civserverport URL parameter rather than the meta
server. This allows many web client instances to connect to the same single
player server. It also allows connecting when the meta server refuses to
start. A hack mode game supports the URL parameters username, savegame,
ruleset and autostart. This makes it possible to bookmark a test case for
quick access.

Hack mode is written in browser side JavaScript. It should therefore not
cause any new security problems.

### DIFF
--- a/freeciv-web/src/main/webapp/javascript/civclient.js
+++ b/freeciv-web/src/main/webapp/javascript/civclient.js
@@ -226,6 +226,37 @@ function init_common_intro_dialog() {
       show_intro_dialog("Welcome to Freeciv-web", msg);
     }
 
+  } else if ($.getUrlVar('action') == "hack") {
+    var hack_port;
+    var hack_username;
+
+    if ($.getUrlVar('civserverport') != null) {
+      hack_port = $.getUrlVar('civserverport');
+    } else {
+      show_intro_dialog("Welcome to Freeciv-web",
+        "Hack mode disabled because civserverport wasn't specified. "
+        + "Falling back to regular mode.");
+      return;
+    }
+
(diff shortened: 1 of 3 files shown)


ghost:
Looks great! You can merge any changes to Freeciv-web that you want.

kvilhaugsvik:
Thank you for the review, Andreas! In this case it was good to get a review, or at least some time without objections, before the merge.

PR #167, opened 2018-06-17
Remove hardcoded server name

lonemadmax:


lonemadmax:
Works for me, but I don't know jsp, so this may not be The Right Way.

ghost:
Looks good

PR #174, opened 2018-07-29
Fix advancement description edge cases

### LINKED ISSUE #173: Advancement description has extra text.
Advancement descriptions that are dead-end (i.e. Fusion Power) have extra "allow" in the description.

With no units, buildings, or dependent tech allowed by an advancement, the infobox reads:
> [Fusion Power](https://en.wikipedia.org/wiki/Fusion_power) (4078) allow.

Whereas it should read:

> [Fusion Power](https://en.wikipedia.org/wiki/Fusion_power) (4078).

mchenryc:
Fixes #173 

This commit uses a number of modern JS features that are not common in the codebase, though examples of all can be found elsewhere. LMK if this is an issue, or a specific browser needs to be targeted.

### DIFF
--- a/freeciv-web/src/main/webapp/javascript/tech.js
+++ b/freeciv-web/src/main/webapp/javascript/tech.js
@@ -410,45 +410,35 @@ function update_tech_screen()
 }
 
 /**************************************************************************
- Returns for example "Bronze working allows building phalax".
+ Returns for example "Bronze working allows building phalanx."
 **************************************************************************/
 function get_advances_text(tech_id)
 {
-  var ptech = techs[tech_id];
+  const num = (value) => value === null ? 'null' : value;
+  const tech_span = (name, unit_id, impr_id, title) =>
+    `<span ${title ? `title='${title}'` : ''}`
+    + ` onclick='show_tech_info_dialog("${name}", ${num(unit_id)}, ${num(impr_id)})'>${name}</span>`;
 
-  var adv_txt = "<span onclick='show_tech_info_dialog(\"" +ptech['name'] + "\", null, null);'>" + ptech['name'] + "</span> (" + Math.floor(ptech['cost']) + ") allows ";
-  var prunits = get_units_from_tech(tech_id);
-  var pimprovements = get_improvements_from_tech(tech_id);
+  const is_valid_and_required = (next_tech_id) =>
+    reqtree.hasOwnProperty(next_tech_id) && is_tech_req_for_tech(tech_id, next_tech_id);
 
-  for (var i = 0; i < prunits.length; i++) {
-    if (i == 0) adv_txt += "building unit ";
-    var punit = prunits[i];
-    adv_txt += "<span title='" + punit['helptext'] 
-            + "' onclick='show_tech_info_dialog(\"" + punit['name'] + "\", " + punit['id'] + ", null)'>" 
-            + punit['name'] + "</span>, ";
-  }
+  const format_list_with_intro = (intro, list) =>
+    (list = list.filter(Boolean)).length ? (intro + ' ' + list.join(', ')) : '';
 
-  for (var i = 0; i < pimprovements.length; i++) {
-    if (i == 0) adv_txt += "building ";
-    var pimpr = pimprovements[i];
-    adv_txt += "<span title='" + pimpr['helptext'] 
-            + "' onclick='show_tech_info_dialog(\"" + pimpr['name'] + "\", null, " + pimpr['id'] + ")'>" + pimpr['name'] + "</span>, ";
-  }
+  const ptech = techs[tech_id];
 
-
-  for (var next_tech_id in techs) {
-    var ntech = techs[next_tech_id];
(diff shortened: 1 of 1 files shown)


lonemadmax:
Great attention to detail, thank you!

But, as much as some may want to ditch MSIE, it's still haunting us. So no ES6, sorry.

andreasrosdalw:
Would it be possible to use ES6, ES7, ES2018 or later in the JavaScript source code in Freeciv-web,
and then compile and minify that using a JavaScript compiler to a version of JavaScript which is widely supported by all browsers?

Freeciv-web currently uses Google Closure Compiler though this Maven plugin:
https://github.com/samaxes/minify-maven-plugin

Perhaps there is a better solution to JavaScript compilation which we can use?
There is for example https://github.com/babel/minify

mchenryc:
Thanks @lonemadmax, I'm actually glad the discussion has come forth, so a clear path can be defined. As mentioned, this PR uses no JS features that are not in freeciv-web already—if our target is legacy, we need to either rollback existing commits, or, as @andreasrosdal points out, add a compiler such as https://babeljs.io/ to the build tool-chain.

Allow me to make a case for moving forward:

To be honest, I question the need to support old MSIE, and if we must, I strongly favor adopting an improved toolchain over reverting existing code.

Can we state clearly *why* MSIE haunts us still? At the time of this comment:
* [Over 2.5 years](https://www.microsoft.com/en-us/windowsforbusiness/end-of-ie-support) have passed since MS stopped supporting old browsers.
* There are no complaints requesting support of legacy browsers (per review of the [56 open, 57 closed issues](https://github.com/freeciv/freeciv-web/issues?q=is%3Aissue) at this time).
* Forbidding the use of modern language features dissuade new contributors.

Indeed, from over a year ago: [You might not need to transpile your JavaScript](https://medium.freecodecamp.org/you-might-not-need-to-transpile-your-javascript-4d5e0a438ca)

Might I recommend: keeping the existing code, allowing modern language features, and prioritizing toolchain support based on need?

lonemadmax:
> Might I recommend: keeping the existing code, allowing modern language features, and prioritizing toolchain support based on need?

Wise advice, indeed.

To be clear, I personally hate MSIE. Probably undeserved hatred from a distant past, but you know, it's difficult to get rid of that. But that is a personal thing, and, even though I have commit access, this is not my project

To be completely clear, I don't know ES6, or other modern stuff. I guess I'll take this opportunity to learn new things, but it won't be something quick.

And I'm not knowledgeable in quite a lot of other things, tooling being just one of them.

With that out of the way...

I'm sorry, I took your comment to mean that those were common features used "out there", that everybody deem safe to use nowadays, not as "in other places in freeciv-web".

I had seen at least `=>` in webgl-inspector, but if I'm right that's only loaded for webgl debugging when the user manually adds a parameter to the URL.

But I was also wrong there. They are also in update_3d_model_live.js and preload.js, and those go to the webclient.js. The arrows are not in the minimized output, so... Yes, it seems [closure is transpiling](https://github.com/google/closure-compiler/wiki/ECMAScript6). Full ES6 compatibility doesn't seem to be a goal, but this PR might pass as is. Const, arrow functions and template functions are on the list, anything else to look for? (Again, I don't know modern features)

About MSIE support, MS had also stopped support for Windows XP in 2014, and in the end they "had to" distribute a patch for Wannacry in 2017. We are stubborn, we continue using what we have.

There are still MSIE users. A patch I sent some time ago had to be changed after applying because trackJs was reporting runtime errors for MSIE users (it wasn't some modern feature, just something that Firefox and Chrome have but MSIE doesn't, but the point is there were real users of MSIE). I don't think there are a lot of them (I'd guess without any data they don't get to 3%), but even if it was just one, I'd rather ask him personally if he didn't mind changing to another browser. It's not like Freeciv-web has hundreds of thousands of users.

With the transpilers that's moot anyway.

The article you link to can be summarized in this context as "You can use modern features and not transpile if you target browsers that support those modern features. And, good news, most of them do". Well, yes, if you don't mind about those 20 users of MSIE, you don't need to care about MSIE at all.

There is an interesting link there: https://philipwalton.com/articles/deploying-es2015-code-in-production-today/. It tells those of us who don't know how to generate two versions, and deliver the clean one to most people while still giving the transpiled one to those that need it.

mchenryc:
Thanks @lonemadmax, for your well thought out response. Also for merging. If you don't mind, I'll continue the thread, briefly, while we distill it into actionable items from the discussion.

I appreciate the backstory on MSIE user feedback through trackJs, which is interesting, and a valid means of determining requirements. Also: #175.

> ...I personally hate MSIE. Probably undeserved hatred from a distant past...

It's OK to say you're not alone in that, right? :wink: 

Then again, there are still enough features they don't support, such that one must polyfill even to use Promises. I rather like the Walton article you linked, so (with apologies) will be submitting a few issues to track both real problems surrounding legacy support (#176), and a spike or two I'd like to attempt around implementing the strategy Walton outlines.

PR #180, opened 2018-08-10
Add empty handler for new ruleset clause packet

zekoz:


lonemadmax:
I would expect this kind of change to be hand in hand with server updates. Right now this is dead code, not even an empty stub. Is anyone reviewing the [changes since 8e2801d583e1aaaee2bfaf3229e2ba233b518ff3 (June)](https://github.com/freeciv/freeciv/compare/8e2801d583e1aaaee2bfaf3229e2ba233b518ff3...master) and how they might affect Freeciv-web?

zekoz:
I'm not sure why you closed, but a handler is definitely needed... https://github.com/freeciv/freeciv/blob/fa63c2a6b39d0d3ff63683fda2268aace248d1c3/server/ruleset.c#L7828

lonemadmax:
I thought it was clear in the other message: a handler is needed only with a server that sends that packet, which was introduced less than a month ago (https://github.com/freeciv/freeciv/commit/fa63c2a6b39d0d3ff63683fda2268aace248d1c3#diff-a56f2ec537afebac742457d2ba401e47), and we are using a version from June: https://github.com/freeciv/freeciv-web/blob/0491f9d0a2602399cb8eda55a2e7844d129ca6ab/freeciv/version.txt#L4

zekoz:
freecivweb.org uses a more recent version that actually sends this packet. Please just merge.

lonemadmax:
And I don't see that in the commit, or the fork.

This change should come with a change in version.txt to point at least to the one that introduces the packet, and that be just one of many commits moving from the current one to there. Do you know freeciv enough and have checked that the rest of changes don't impact freeciv-web in any way (I'm not attacking you or asking for a curriculum, I myself can't do it except for the most simple ones)?

That said, merging on the basis that it is dead code that will have to be included some day and doing it now seems it may help freecivweb.org a bit.

But it's probably my last of this kind. I won't disagree with others doing it, though; they'll have to step in anyway for any change in the freeciv tracked commit.

zekoz:
freecivweb.org now runs based on d196b8a0131637ce5947c15d04231c493bff03fb. You should now be able to change version.txt to refer to that.

kvilhaugsvik:
I just started updating from the current revision. I'll hopefully be able to test and merge the update during the weekend. (Any new dependencies, configuration options or other stuff I'll have to change to make my bare metal install from June work?)

mchenryc:
@kvilhaugsvik 
> ... or other stuff I'll have to change...

If you are not using a clean checkout, you may have old files in `freeciv-web/src` that need to be manually deleted (per changes in #186, see [this comment](https://github.com/freeciv/freeciv-web/pull/186#issuecomment-416828204)). Git should mark them as untracked, but I suspect anyone used to the old build may expect them to be present.

PR #185, opened 2018-08-18
Clean up install/provision output

mchenryc:
Still a lot of noise, but this removes the 'progress updates' of some commands, and unnecessary extraction logging.

apt-get change from: https://stackoverflow.com/a/25907959/80714

### DIFF
--- a/freeciv-web/build.sh
+++ b/freeciv-web/build.sh
@@ -18,4 +18,4 @@ else
 fi
 
 echo "maven package"
-mvn flyway:migrate package && cp target/freeciv-web.war "${TOMCATDIR}/webapps/"
+mvn -B flyway:migrate package && cp target/freeciv-web.war "${TOMCATDIR}/webapps/"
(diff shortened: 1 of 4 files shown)


lonemadmax:
Any reason to make freeciv/freeciv-web.project executable?

I'll miss the colors in mvn output when rebuilding, but I'll trade them to getting rid of the progress messages during install.

mchenryc:
> I'll miss the colors...

When we can upgrade to maven to to 3.5.3, colors "should" be available, according to: https://issues.apache.org/jira/browse/MNG-6296

lonemadmax (changes requested):
When possible, please revert the change in  freeciv/freeciv-web.project permissions.

You may also want to change the comment: the SO link is for the mvn change. But the apt-get one is easy to find now that you did the work and we can directly search the change.

mchenryc:
Thanks for the pr-proofreading - All addressed.

mchenryc:
@lonemadmax FWIW, you're notice of the execution bit flip prompted me to find [this](https://stackoverflow.com/a/1580644/80714), which solves something that has plagued me for years:

    git config --global core.fileMode false

Perhaps not "recommended" but I find it immediately useful.

Issue #176, opened 2018-08-05
JS code not being transpiled?


mchenryc:
Freeciv-web needs to be transpiled to run on MSIE, but when I run a fresh build from master on MSIE-11.0.9600.19080, I see:

> SCRIPT1014: Invalid character
> File: webclient.js, Line: 31831, Column: 28

which points to: 
```
  model.traverse((node) => {
```
...an IE6 arrow-function, which [Closure should transpile](https://github.com/google/closure-compiler/wiki/ECMAScript6#transpilation).

Hopefully this is just a configuration issue, or perhaps I'm building incorrectly somehow.

mchenryc:
At minimal, closure needs to be updated, as many improvements [have been made](https://github.com/google/closure-compiler/wiki/Releases), many to handle ES6 features, since the currently used v20170521 release.

andreasrosdalw:
https://github.com/samaxes/minify-maven-plugin/pull/155

Perhaps you can try adding support for transpiling using https://babeljs.io ?

lonemadmax:
That's the raw code for local testing. The "production" host would serve webclient.min.js: https://github.com/freeciv/freeciv-web/commit/a9cc731a561cd8f44c32bc319b4c2b14fa51e53b#diff-8d3042c0d730641af06832fd05e11539, that is transpiled.

I'm keeping this open with the more general issue of sending modern JS to modern browsers, as discussed in #174.

mchenryc:
**TL;DR:** I was looking at the right file - but hadn't realized the non-minified file is downloaded when hosted locally - IE was failing no matter what I did. I then began investigating closure-compiler's latest version, which does _not_ transpile when using certain settings. Closure-compiler v20180716, using `<closureLanguageIn>ECMASCRIPT_2017</...>` works fine (as do our current settings w/ current version of CC).

**Takeaway**: 
1. We shouldn't test hostname to decide what script to load, but rather against some config setting specific to which build configuration should be used. This is related to #175 (in that it tests against hostname as well).

    <% if (request.getServerName().equals(fcwHost)) { %>

2. I'll submit a patch for updating to closure-compiler's latest release. Sounds like there are a number of bugfixes.

**The tangential rabbit hole** (for posterity): upgrading closure-compiler for [their bug-fixes](https://github.com/google/closure-compiler/wiki/Releases):

[...]

PR #186, opened 2018-08-19
Keep src dirs clean

### LINKED ISSUE #182: Keep source directories clean of build artifacts
During the build, many files are generated directly into the same directory as source controlled. This adds noise to the repository, making a "clean build" error prone, and risks the commit of one of these generated files.

These are mostly derived from upstream freeciv, via python and shell scripts during the general install (see [scripts/install/install.sh](https://github.com/freeciv/freeciv-web/blob/develop/scripts/install/install.sh)), which doesn't play particularly well with maven expectations.

Many generated files are source files which need to be known to IDEs (and thus "easier" to work with when generated to source directories), while others are images, HTML help files, and other resources that are just as easily generated to a target/build/dest directory.

The hope then, is that after running an install/build, the following will hold true:
* `git status` will show no changed or untracked files after full build, and test.
* maintain simplicity of overall build/dev lifecycle with Vagrant
* incremental builds of the webapp will (continue to) not need to regenerate derived files

mchenryc:
Fixes #182 

While not intended to modify so much of the bash/python scripts, I did some normalizing and clean up. Review will be easier by commit.

With this, the downloaded `tests/casperjs-xxxx` appears to be the remaining untracked item (directory) after a build, though I am leaving that to be fixed after tests are updated. A `pbem-games.json` file [was mentioned](https://github.com/freeciv/freeciv-web/issues/182#issuecomment-412368454), though that doesn't seem to be generated by a standard install.


andreasrosdalw:
Great improvement. However, Travis build failed. Have you tested the Vagrant build on Windows?

lonemadmax:
The error seems to be that in `scripts/sync-js-hand.sh` `gen_event_types.py` is now called directly instead of via python interpreter, and it is not executable. `update-wikipedia-docs.py` may suffer from the same problem.

You may need to run at least one pbem game before the `pbem-games.json` is created. Maybe it should be saved somewhere else or even to the DB, instead of ignored.

mchenryc:
Strange that the interpreter fails - does Travis run tests on something other than the vagrant box which might have different config? Multiple runs on vagrant locally ensure that the shebang changes were all working (as they say "for me"...)

Investigating.

mchenryc:
BTW, the execution bit touches on the windows/virtualbox disparity - I do run on windows, and can't right now use git from the virtual-box because ~~checking out via windows got me CRLF line endings on local text files~~ (actually that was an assumption, and a closer look indicates it might be just execution bit changes, which is/could be due to git/bash, or cygwin, or... regardless, it's a mess), so git thinks everything has changed from the Ubuntu side :man_facepalming: .

It looks like this has been a problem for freeciv-web in the past, with the .gitattributes, and the various sed calls during install, forcing them to be LF.

IMHO, it would be great if we could just force all text files to be LF, and force all commits to be "fixed" silently, even if there are copy paste errors from... notepad or whatever (my editors are fine w/ LF on windows, but clearly, not all are!). To that end, I found this: https://stackoverflow.com/a/42135910/80714 (ok, many others too) where the OP has the same situation (VirtualBox hosted on Windows, running Linux, and wants to work from both sides) which purport to "fix it, once and for all".

@andreasrosdal or @lonemadmax, care to experiment with this w/ me? I expect the `* text=auto eol=lf` git magic might work on a local machine, but wonder how such a change might affect others that handn't run the cache-clearing or resetting commands. (and it would actually please me to remove the sed commands that just fix line endings... if we can trust that they're not necessary :grin:)

mchenryc:
> You may need to run at least one pbem game before the pbem-games.json is created. Maybe it should be saved somewhere else or even to the DB, instead of ignored.

I was able to run the test, which generated the file. It's now ignored as well, though if there's going to be an effort to put the pbem saved games into the DB... let's make that a separate issue.

Perhaps of note: I have a general preference for putting `.gitignore` files in the directory of the files that are being ignored (as here), in the "principal of least privilege" vein.

lonemadmax:
> care to experiment with this w/ me? I expect the * text=auto eol=lf git magic might work on a local machine, but wonder how such a change might affect others that handn't run the cache-clearing or resetting commands

I'm not sure what to test. In case you need to see how it goes when pulling a commit with the change in .gitattributes, I've made it to the fix-line-endings branch.

I've played myself in my linux box. It doesn't make a difference for me, but I guess that was expected.

There are a few CRLF files already:
```
publite2/pubscript_pbem.serv
pbem/pbem.py
freeciv-web/src/main/webapp/javascript/libs/range.js
freeciv-web/src/main/webapp/javascript/libs/bmp_lib.js
freeciv-web/src/main/webapp/javascript/libs/timer.js
freeciv-web/src/main/webapp/javascript/libs/slider.js
freeciv-web/src/main/webapp/javascript/scorelog.js
freeciv-web/src/main/webapp/javascript/pbem.js
freeciv-web/src/main/webapp/css/bluecurve.css
```

`git status` doesn't show them as changed.
`git checkout` doesn't automatically change them to LF.
`git diff` doesn't show differences.
After changing a line (just a random character, not line endings) `git diff` shows just that line, with `^M` mark.
A checkout gets it again with CRLF as it was.
Changing a line and adding the file to the index still changes just that line.
Committing still keeps CRLF.
`rm` and `reset` don't change anything.
`git add --renormalize .` does add all of them to the index.
Committing still keeps them as CRLF in working dir, but `git diff` after a change shows it without `^M`, and changing just line endings doesn't count as a change, with both `git status` and `git diff` showing nothing, and with a real change there's a warning saying that CRLF will be changed to LF in the repo but kept as CRLF in your working dir.
After that the `rm` and `reset` commands do change them all to LF in working dir.

Branch fix-line-endings-norm has the change in `.gitattributes` and the renormalization of those other files in one commit.

Branch fix-line-endings-norm-change builds on that and adds a change to `scripts/install/systems` (both a line change and a new line), a file that probably has CRLF in windows, if you want to check whether it has CRLF, LF or both after that commit without having done the `rm` and `reset` magic.

mchenryc:
> After changing a line (just a random character, not line endings) git diff shows just that line, with ^M mark.

Git diff will show ^M for changed lines with CRLF, but not on un-changed adjacent lines, even if they are CRLF as well, (it is flagging only changed lines with errors). Somewhat misleading. See `core.whitespace` in git config](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration).

To your branches:
* `fix-line-endings` still has CRLF in the files you mention, as well as `.gitattributes`, as expected.
* `fix-line-endings-norm` does not have any CRLF files when I check it out.
* `fix-line-endings-norm-change` does not have CRLF files either.

The use of `.gitattributes` and doing the reset seems to be effective. `scripts/install/systems` had LF endings only.

I've modified it again, changing _all_ lines to CRLF and added `NEW_FILE.md` with all CRLF. Adding that to the repo got the warning "CRLF will be replaced by LF...". The push to my repo has ["fixed" `scripts/install/systems`](https://github.com/freeciv/freeciv-web/compare/fix-line-endings-norm-change...mchenryc:fix-line-endings-norm-rechange) back to LF.

I'd suggest incorporating your `.gitattributes` change with the EOL corrections. Never mind deleting the `sed` lines of course - I see they are to protect from local edits.

I'll consider this PR complete unless there are further fixes requested.

lonemadmax:
Sorry I have not started with this before. Here are a few small issues/notes:
- Some data files have gained executable bits:
```
freeciv-web/src/derived/.gitignore
freeciv-web/src/derived/README.md
freeciv/freeciv-web.project
pbem/.gitignore
```
- The first parameter in python's argparse is the program name (optional, the default being... the program name). Change the calls to `ArgumentParser(description='...')`. There's also some copy-paste in there, with the description of the wikipedia extraction in:
```
scripts/generate_js_hand/generate_js_hand.py
scripts/helpdata_gen/helpdata_gen.py
scripts/soundspec-extract/soundspec-extract.py
```
The verbose option help in `scripts/freeciv-img-extract/img-extract.py` is also a copy-paste remnant.
- Maybe delete `TEMP_DIR` at the end of `scripts/freeciv-img-extract/img-extract.py` if everything's gone right? Not that this is a must, it should disappear automatically some time in the future...
- I don't know how up-to-date `scripts/autodeploy.sh` is, but maybe its calls to sync/sync-js-hand should be updated.
- Now some flags are not converted to webp. `convert` returns with nonzero exit status, but the png is generated.
- In an already installed system a new build asks to regenerate the derived files as they're not found (nice), but it also takes the old ones. Users must remove the old files or they will have mixed stale stuff. Maybe a notice in the merge commit will be enough. Or maybe we can inaugurate `scripts/migration`, although that would still be just a notice.
- The problem with jsSourceIncludes in pom.xml seems to be that it doesn't go up jsSourceDir. It can be workarounded by setting a higher jsSourceDir and updating all the file references. You wouldn't have to declare each of the derived files, but I find it ugly and dirty. Like this (or one more level to include generated-sources):
```
<jsSourceDir>../..</jsSourceDir>
<jsFinalFile>target/freeciv-web/javascript/webclient.js</jsFinalFile>
<jsSourceFiles>
	<!-- order is important: these must be before sources below -->
	<jsSourceFile>main/webapp/javascript/libs/EventAggregator.js</jsSourceFile>
	<jsSourceFile>main/webapp/javascript/libs/handlebars.runtime-v4.0.11.js</jsSourceFile>
	<jsSourceFile>../target/generated-sources/javascript/hbs-templates.js</jsSourceFile>
</jsSourceFiles>
<jsSourceIncludes>
	<jsSourceInclude>main/webapp/javascript/*.js</jsSourceInclude>
	<jsSourceInclude>main/webapp/javascript/libs/*.js</jsSourceInclude>
	<jsSourceInclude>main/webapp/javascript/2dcanvas/*.js</jsSourceInclude>
	<jsSourceInclude>main/webapp/javascript/webgl/*.js</jsSourceInclude>
	<jsSourceInclude>derived/webapp/javascript/*.js</jsSourceInclude>
	<jsSourceInclude>derived/webapp/javascript/2dcanvas/*.js</jsSourceInclude>
</jsSourceIncludes>
<jsSourceExcludes>
	<jsSourceExclude>main/webapp/javascript/frontpage.js</jsSourceExclude>
	<jsSourceExclude>main/webapp/javascript/libs/jquery.min.js</jsSourceExclude>
	<jsSourceExclude>main/webapp/javascript/libs/hammer.min.js</jsSourceExclude>
	<jsSourceExclude>main/webapp/javascript/tileset_config*</jsSourceExclude>
	<jsSourceExclude>main/webapp/javascript/tileset_spec*</jsSourceExclude>
	<jsSourceExclude>main/webapp/javascript/libs/gif.worker.js</jsSourceExclude>
</jsSourceExcludes>
```

mchenryc:
> Sorry I have not started with this before

You're reviews are thorough, and most welcome; thanks for taking the time at all. My apologies if I am becoming lax, relying on you to catch small errors.

**WRT `autodeploy.sh`** I've updated it, but notice it is 1) `travis` specific, and 2) tries to pull from `master` branch (not our default: `develop`). Anyone (...@andreasrosdal) know if this makes autodeploy obsolete? The two share a lot of duplication (and yes, I've now added more...), perhaps another 'issue' should be generated to investigate, and DRY up the two, remove `autodeploy.sh` if it is dead, or combine `autodeploy.sh` and `travis-run.sh`.

Issues:
* [x] execution bits
* [x] `ArgumentParser` descriptions
* [x] remove temp dir - I considered this as well, but opted to leave it for possible debugging purposes (given these temp files had not been deleted previously either). I've left it as is for now
* [x] `scripts/autodeploy.sh` updated
* [x] flags conversion errors - .webp files are now generated even if .png creation had errors, as long as .png file is created
* [x] handling of files generated-from-external sources
I sense leaving the old is a good thing, given they are being asked to review all for suitability. I've added a further notice of this upon success. Will this work?
* [x] freeciv-web js compilation
I'm inclined to leave this as is, continuing js build changes in the node migration discussed in #176. The spike using node (keeping current dir-structure) has both advantages and disadvantages, and conflicts enough with this PR that it would be worthwhile to continue discussion there with a new PR, after this one lands.

mchenryc:
Also added a solution to restore maven output colors, lost since https://github.com/freeciv/freeciv-web/pull/185#issuecomment-414139796

andreasrosdalw:
autodeploy is not in use any more. It was used for autodeploys on play.freeciv.org, these days I deploy manually, since it's safer.

lonemadmax:
It seems you've only forgotten [the text for the verbose option in img-extract](https://github.com/freeciv/freeciv-web/pull/186/files#diff-ed9a7e16cc22a08e96c8e91a8520af44R29) and naming the argparse parameter. Compare:
```
parser = argparse.ArgumentParser('My description')

$ ./test.py 
usage: My description [-h] -f FREECIV [-o OUTDIR]
My description: error: the following arguments are required: -f/--freeciv
$ ./test.py -h
usage: My description [-h] -f FREECIV [-o OUTDIR]

optional arguments:
  -h, --help            show this help message and exit
...
```
With:
```
parser = argparse.ArgumentParser(description='My description')

$ ./test.py 
usage: test.py [-h] -f FREECIV [-o OUTDIR]
test.py: error: the following arguments are required: -f/--freeciv
$ ./test.py -h
usage: test.py [-h] -f FREECIV [-o OUTDIR]

My description

optional arguments:
  -h, --help            show this help message and exit
...
```

Re leaving old files, I was not talking about the ones from wikipedia, sorry for the confusion. I meant the ones this PR is about, especially the js in src. Say you have a server, now with `freeciv-web/src/main/webapp/javascript/packhand_gen.js` and company. You get these changes and then some more, which include a freeciv update. Regenerate the files and `webclient.js` gains both old `packhand_gen.js` from `main` and the new one from `derived`. Not only that, but the new one is forced to the top, so the old one overwrites it. For most things it won't matter and some variables and functions will be defined twice with the same value, but there may be not only additions but also changes (say a new packet type in `client_handle_packet`), and the old versions will win.

Not this PR's fault, of course. I'm just wondering if my adding a notice in the merge commit will be enough.

mchenryc:
Python scripts updated, thanks.

Ah, yes, I missed your point; the .js files generated by the old "sync" scripts (directly into the `src/webapp/...` directories) will still be there, for those who have previously built the project (_precisely_ the problem with ever generating files into the src directory... :slightly_frowning_face: )\*.

options include...
1. Tell devs/users to delete the files that aren't part of the build. It would be a one time thing (once local branches are rebased onto `develop`). "How to tell" them is a problem, and silently including the old files is also a potential problem, as you point out.
2. Delete the old files automatically. I have a rather negative reaction to having the build script modifying anything in the `src` dir, especially deleting source files.
3. Refuse to build if the files exist, informing devs why, and how to continue.

I don't think **1** is that bad. `git status` complains profusely for all these untracked files, and I for one find that motivation enough to delete them.\*\* I will keep my own note files around, but in general, git status should remain clean. We would want to ensure that Travis does a clean checkout as well.

My next preference would be **3**. It only needs to happen once (unless one often switches between un-rebased branches and `develop`), it actively prevents the problem, forces the deletion to be explicit, and is silent once taken care of. It won't be an issue for devs on fresh-checkout, so such a break won't be a hurdle for newcomers.

Regardless, once merged, branches should have this change merged in as soon as possible - switching back and forth is tedious.

\* The `src/derived` solution isn't perfect either, as generated files are still in `src`, though they are relegated to a single directory which one can `rm -rf src/derived/webapp` for a pristine workspace.

\*\* In honesty, keeping git output clean was a major motivation for this PR. Sure one could just `.gitignore` them away, but that's just hiding the problem.

PR #187, opened 2018-08-20
Keep trackjs token in config, like google signin key

### LINKED ISSUE #175: Trackjs token should be removed from source
The trackjs token is currently in the committed code. Unless this is publicly available, the token should be part of the server's environment variables, read by the servlet, and injected as a global JS constant. The trackjs `<script...` tags should be protected by testing, and use this variable, not by comparing the current servername to `fcwHost`.

OTOH, if user-metrics are publicly available, cool - but we should include information on how to access them.

If not public, I would also recommend generating a new token (and permanently disabling the old through whatever means trackjs provides), rather than [rewriting](https://help.github.com/articles/removing-sensitive-data-from-a-repository/) [history](https://medium.com/bam-tech/remove-sensitive-information-from-your-git-repository-10cb421f1b84), to erase it from git.

mchenryc:
Fixes #175 

* trackjs scripts are only loaded if `trackjs-token` is specified
* trackjs-token is empty (disabled) by default, and not tied to
  a hostname. Uncomment and specify in config if you want tracking
* Also adds abilty to show un-minified `webclient.js` and `webclient.css`
  by adding `?debug` to url, even on hosted sites (still uses fcw_host
  to send minified)

### DIFF
--- a/freeciv-web/src/main/webapp/WEB-INF/config.properties.dist
+++ b/freeciv-web/src/main/webapp/WEB-INF/config.properties.dist
@@ -3,6 +3,7 @@ savegame_dir=/var/lib/tomcat8/webapps/data/savegames/
 email_username=username
 email_password=password
 google-signin-client-key=122428231951-2vrvrtd9sc2v9nktemclkvc2t187jkr6.apps.googleusercontent.com
+#trackjs-token=ee5dba6fe2e048f79b422157b450947b
 # Used in some messages and to distribute minimized or expanded js files,
 # so put the production server even when testing.
 fcw_host=example.com
--- a/freeciv-web/src/main/webapp/WEB-INF/config.properties.dist
+++ b/freeciv-web/src/main/webapp/WEB-INF/config.properties.dist
@@ -4,6 +4,5 @@ email_username=username
 email_password=password
 google-signin-client-key=122428231951-2vrvrtd9sc2v9nktemclkvc2t187jkr6.apps.googleusercontent.com
 #trackjs-token=ee5dba6fe2e048f79b422157b450947b
-# Used in some messages and to distribute minimized or expanded js files,
-# so put the production server even when testing.
+# Used in some messages, so put the production server even when testing.
 fcw_host=example.com
--- a/freeciv-web/src/main/webapp/webclient/index.jsp
+++ b/freeciv-web/src/main/webapp/webclient/index.jsp
@@ -6,19 +6,20 @@
 <%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
 <%
 String googleSigninClientKey = null;
-String fcwHost = null;
 String trackJsToken = null;
+boolean fcwDebug = false;
+String fcwMinified = "";
 try {
   Properties prop = new Properties();
   prop.load(getServletContext().getResourceAsStream("/WEB-INF/config.properties"));
   googleSigninClientKey = stripToEmpty(prop.getProperty("google-signin-client-key"));
   trackJsToken = stripToNull(prop.getProperty("trackjs-token"));
-  fcwHost = prop.getProperty("fcw_host");
+
+  String debugParam = request.getParameter("debug");
+  fcwDebug = (debugParam != null && (debugParam.isEmpty() || parseBoolean(debugParam)));
+  fcwMinified = fcwDebug ? "" : ".min";
 } catch (IOException e) {
   e.printStackTrace();
-}
-if (fcwHost == null || fcwHost.isEmpty()) {
-  fcwHost = request.getServerName();
 }
 %>
 <!DOCTYPE html>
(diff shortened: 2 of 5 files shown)


freeciv-web/src/main/webapp/WEB-INF/config.properties.dist
```diff
@@ -3,6 +3,7 @@ savegame_dir=/var/lib/tomcat8/webapps/data/savegames/
 email_username=username
 email_password=password
 google-signin-client-key=122428231951-2vrvrtd9sc2v9nktemclkvc2t187jkr6.apps.googleusercontent.com
+#trackjs-token=ee5dba6fe2e048f79b422157b450947b
```
lonemadmax: https://github.com/freeciv/freeciv-web/blob/3059122cc0145fa966c7452d2287cd76a2bdc083/freeciv-web/src/main/webapp/WEB-INF/jsp/fragments/head.jsp#L30 will need to change too.

freeciv-web/src/main/webapp/webclient/index.jsp
```diff
@@ -1,48 +1,47 @@
...
-  googleSigninClientKey = prop.getProperty("google-signin-client-key");
-  fcwHost = prop.getProperty("fcw_host");
+  googleSigninClientKey = stripToEmpty(prop.getProperty("google-signin-client-key"));
+  trackJsToken = stripToNull(prop.getProperty("trackjs-token"));
+
+  String fcwHost = stripToNull(prop.getProperty("fcw_host"));
+  String debugParam = request.getParameter("debug");
+  boolean hosted = fcwHost != null && request.getServerName().equals(fcwHost);
+  boolean debug = (debugParam != null && (debugParam.isEmpty() || parseBoolean(debugParam)));
+  minified = hosted || !debug ? ".min" : "";
```
lonemadmax: I'm not sure I understand either the comment or the code. This will only show unminified code when the site is not hosted and the request has the debug param. That is, on hosted site it will always be minified and on dev machine you will have to add the debug parameter. Which may be OK, there's value in testing the minified code (which will be the real one in prod), I just don't think that was the intention given the commit comment.

I do like the debug param, BTW. To the point of thinking about not using the host name in this context, or even doing `hosted ^ debug` so that default behaviour is as it was and `debug` just changes it, making both raw and minified runs possible in both hosted and unhosted cases.

freeciv-web/src/main/webapp/webclient/index.jsp
```diff
@@ -1,48 +1,47 @@
...
 <meta name="description" content="Freeciv-Web is a Free and Open Source empire-building strategy game inspired by the history of human civilization.">
-<% if (request.getServerName().equals(fcwHost)) { %>
-<script type="text/javascript">window._trackJs = { token: 'ee5dba6fe2e048f79b422157b450947b' };</script>
+<% if (trackJsToken != null) { %>
+<script type="text/javascript">window._trackJs = { token: '<%= trackJsToken %>' };</script>
 <script type="text/javascript" src="https://cdn.trackjs.com/releases/current/tracker.js"></script>
 <% } %>
 <script type="text/javascript">
 var ts="${initParam.buildTimeStamp}";
-var fcw_host="<%= fcwHost %>";
```
lonemadmax, line 35: OK, but keep something with the result of the new test and update:
https://github.com/freeciv/freeciv-web/blob/3059122cc0145fa966c7452d2287cd76a2bdc083/freeciv-web/src/main/webapp/javascript/webgl/mapview.js#L108

https://github.com/freeciv/freeciv-web/blob/3059122cc0145fa966c7452d2287cd76a2bdc083/freeciv-web/src/main/webapp/javascript/webgl/renderer_main.js#L43

lonemadmax (changes requested):

freeciv-web/src/main/webapp/WEB-INF/config.properties.dist
mchenryc: Fixed. I've included the same parameter initialization as the the page that doesn't use head.jsp, and also used the same `window._trackjs`, as this is what trackjs recommends.

freeciv-web/src/main/webapp/webclient/index.jsp
mchenryc: There was a defect in the initial version wrt debug. If you're good relying on just debug though, I rather prefer that, as it makes everything consistent.

freeciv-web/src/main/webapp/webclient/index.jsp
mchenryc: pending

mchenryc: Fixed

freeciv-web/src/main/webapp/webclient/index.jsp
```diff
@@ -6,45 +6,43 @@
...
 try {
   Properties prop = new Properties();
   prop.load(getServletContext().getResourceAsStream("/WEB-INF/config.properties"));
   googleSigninClientKey = stripToEmpty(prop.getProperty("google-signin-client-key"));
   trackJsToken = stripToNull(prop.getProperty("trackjs-token"));
-  fcwHost = prop.getProperty("fcw_host");
+
+  String debugParam = request.getParameter("debug");
+  fcwDebug = (debugParam != null && (debugParam.isEmpty() || parseBoolean(debugParam)));
+  fcwMinified = fcwDebug ? ".min" : "";
```
lonemadmax: You surely meant the opposite: debug -> raw, !debug -> min.

mchenryc: That is a distinct possibility :smirk:.

Fixed.

lonemadmax (approved):

PR #191, opened 2018-08-28
Allow numbers in usernames

lonemadmax:


freeciv-web/src/main/java/org/freeciv/services/Validation.java
```diff
@@ -0,0 +1,42 @@
...
+ *******************************************************************************/
+package org.freeciv.services;
+
+import java.util.regex.Pattern;
+
+public class Validation {
+
+	private static Pattern usernameInvalidPattern = Pattern.compile("[^a-z]");
+
+	public static boolean isValidUsername(String name) {
```
mchenryc: This should be an instance method, not `static`. Static methods are difficult to test, both the static method, and the classes that use it (although the classes that use it can't be easily tested independent of the service, unless DI is used). Ideally we would use a DI framework (e.g. Guice which is more lightweight than Spring), but until then, creating an instance in each class that uses it has advantages.

Games, and Statistics services use instance methods as well.

freeciv-web/src/main/java/org/freeciv/services/Validation.java
```diff
@@ -0,0 +1,42 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *******************************************************************************/
+package org.freeciv.services;
+
+import java.util.regex.Pattern;
+
+public class Validation {
+
+	private static Pattern usernameInvalidPattern = Pattern.compile("[^a-z]");
```
mchenryc: Should be `final`.

freeciv-web/src/main/java/org/freeciv/servlet/DeactivateUser.java
```diff
@@ -43,7 +45,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
 		String username = request.getParameter("username");
 		String secure_password = java.net.URLDecoder.decode(request.getParameter("sha_password"), "UTF-8");
 
-		if (username == null || username.length() <= 2) {
+		if (!Validation.isValidUsername(username)) {
```
mchenryc: (see comments in Validation)

Unlike the way Games and Statistics are used, I would suggest creating a field in this class for `private final Validation validation ...` (of course this applies to all uses of Validation in the PR). I would recommend assigning it in the constructor - that is better than creating one in each method like Games/Statistics, because HttpServlet is a long lived object, the services are stateless and thread-safe (as they should be), so a single instance can be used for all requests.

Note that Games/Statistics are also safe to be used in this way.

This won't help testing (we still won't be able to inject a mock version of the service without using other patterns), but what I've described is a very easy to those patterns or DI.
Using a DI framework would allow us to reuse the same service instance between all Servlets (e.g. `Games` is used in both `Index`, and `GameList`)... again though, that's a separate issue.

mchenryc (changes requested):

freeciv-proxy/freeciv-proxy.py
```diff
@@ -188,10 +188,10 @@ def get_civcom(self, username, civserverport, ws_connection):
...
 
 def validate_username(name):
-    if (name is None
-            or len(name) <= 2 or len(name) >= 32
-            or name == "pbem"
-            or re.search('[^a-zA-Z]', name) is not None):
+    if (name is None or len(name) <= 2 or len(name) >= 32):
+        return False
+    name = name.lower()
+    if (name == "pbem" or re.match('\A[a-z][a-z0-9]*\Z', name) is None):
```
mchenryc: Are the `A` and `Z` intentional?

mchenryc: This would probably be more clear simply returning the result of the clause (no `if` statement, and inverted to get the correct boolean result), and deleting the next 3 lines.

mchenryc (changes requested):

freeciv-web/src/main/java/org/freeciv/services/Validation.java
```diff
@@ -0,0 +1,42 @@
...
+ *******************************************************************************/
+package org.freeciv.services;
+
+import java.util.regex.Pattern;
+
+public class Validation {
+
+	private static Pattern usernamePattern = Pattern.compile("[a-z][a-z0-9]*");
+
+	public static boolean isValidUsername(String name) {
```
mchenryc: This should be an instance method, not static. Static methods are difficult to test - both the static method, and the classes that use it (although the classes that use it can't be easily tested independent of the service, unless DI is used). Ideally we would use a DI framework (e.g. Guice which is more lightweight than Spring), but until then, creating an instance in each class that uses it has advantages.

Games, and Statistics services use instance methods as well.

freeciv-web/src/main/java/org/freeciv/services/Validation.java
```diff
@@ -0,0 +1,42 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *******************************************************************************/
+package org.freeciv.services;
+
+import java.util.regex.Pattern;
+
+public class Validation {
+
+	private static Pattern usernamePattern = Pattern.compile("[a-z][a-z0-9]*");
```
mchenryc: Should also be `final`.

mchenryc (changes requested):

freeciv-proxy/freeciv-proxy.py
lonemadmax: Yes, they were intentional. But back to documentation, `\A` is unnecessary. And they could be the more common `^` and `$` as long as the multiline flag is not used. And there's `fullmatch` that matches against the whole string, so I'll remove the anchors and use `fullmatch`.

freeciv-web/src/main/java/org/freeciv/services/Validation.java
lonemadmax: I'm afraid I don't get this one. I've read about the problems of static methods and I don't see how they apply here. This is a pure function, with the class just providing a kind of namespace, without any state, and other future methods should be that way. Maybe I'm shortsighted.

Not having used testing frameworks, I don't understand the difficulty of testing static methods (the really static ones that are functions with no side effects, I mean). You'd just call them with a set of interesting cases for their arguments and compare the result with the expected one. As to testing the users, I can't discuss that, but I don't want an automated test telling me a caller does the right depending on what that method returns; I want it to tell me it does the right thing with usernames with right and wrong formats.

Now, given my inexperience, just a "trust me, you'll regret it in a few months if you make this static" will be enough to change it, but I'd rather understand why, if it can be explained in a simple way to someone who is not a software developer.

freeciv-proxy/freeciv-proxy.py
mchenryc: TIL today - Python regex has things not even Perl has!

freeciv-web/src/main/java/org/freeciv/services/Validation.java
mchenryc: > pure function, with the class just a class providing a kind of namespace, without any state

All true, but not relevant to testability.

`isValidUsername(...)` is simple, but has 5 branches. Yesterday, passing "myname1" would produce `false`, but today `true`. In the future, other changes may be made.

With that in mind, when testing a method (e.g. `MyClass.testMe(...)`) that _uses_ `Validation.isValidUsername(...)`, you want to test cases where `isValidUsername(...)` passes, and where it fails within `testMe(...)` (and maybe when it throws an exception) - so if you _must_ use the real `isValidUsername(...)`, you also must know what input will cause it produce those results. In short, you're testing `testMe(...)`, and implicitly `isValidUsername(...)` also, making the tests more [complicated](http://noop.nl/2008/08/simple-vs-complicated-vs-complex-vs-chaotic.html).

To avoid that, we mock `isValidUsername(...)` - use a fake version of the method that simply does what we want during tests, regardless of input. Changing implementation of a static method [is possible](https://github.com/powermock/powermock/wiki/Motivation) (with behavioral edge-cases due to their "global" nature), but not recommended - change it once you're changing it everywhere. You need to change it _back_ to the original, etc... With instances, a different mock instance is used for each test, avoiding the problems of static methods, and insuring the changes don't affect other tests (or worse, the test framework).

Assuming we make `isValidUsername(...)` an instance method, how does `MyClass` get an instance of `Validation` - as a mock (during tests) or otherwise (at runtime)? The "[Dependency Injection](https://en.wikipedia.org/wiki/Dependency_injection)" pattern (aka. "inversion of control", "IOC", etc). Using DI, during our tests we create a mock `Validation` instance for each test, and for each test of `testMe(...)` tell the `validation` mock how `isValidUsername(...)` should behave.

Consider this JUnit / [Mockito](https://site.mockito.org/) based test, where `testMe(...)` changes state if the argument is valid (worth noting that annotations be used to reduce the code more, but I thought this would be more clear to those unfamiliar w/ mockito):

    public class MyClassTest {
        Validation mockValidation;
        MyClass sut;

        @Before
        void setup() {
            mockValidation = mock(Validation.class);
            sut = new MyClass(mockValidation); // constructor injection (preferred) though setter methods work
        }
        @Test
        void testMe_withValidUsername_changesState() {
            when(mockValidation.isValidUserName(any())).thenReturn(true); // regardless of input, return true

            myClass.testMe("don't care at all");

            assertThat(myClass.getState(), equalTo(expectedStateChanged)); // verify behavior of `testMe`
        }
        // testMe_withInvalidValidUsername_noStateChange...
    }

As you point out, `isValidUsername(name)` is a pure function without state. Image a method  `calculateMyTaxReturnPayment(x,y,z,p,d,q,r,o,f,l,m,a,O)` that is also a pure function and stateless... as is `String.copyValueOf(charArray)`. _Really_ simple utility functions (e.g. language runtime API like the `String` statics), you clearly _can't_ change, and wouldn't want to anyway. Some application utilities are so simple, they don't bear being injected (e.g. `MyStringUtils.firstLetter(...)`). So how to decide what needs to be testable? It's a judgement call, but a good rule of thumb is that if there is _any_ business logic, it should be testable, (username validity is clearly business logic, given it is evolving). I generally err on the side of caution, with the "trust me you'll regret it..." warning.

Finally, in order to use DI, a framework to create all objects in the application and "wire them up" is needed. [Guice](https://github.com/google/guice) is a good, lightweight one, you may have heard of [Spring](https://spring.io/) which has DI (and much more), and there are others. We don't have DI (yet!) in freeciv-web, though, again, erring on the side of caution, I prefer to follow the pattern as much as possible to reduce refactoring later. Thus I suggest:

1. Design the classes as testable, instantiating the service (Validation) in the constructor of classes where used.
2. Incorporate a DI framework, making only small modification to the constructors.
3. Add tests

The second and third could be swapped - or never happen, 1. is still a good choice regardless.

Thanks for listening - apologies for the length

lonemadmax: > apologies for the length

On the contrary, thank you for taking the time and effort to write it.

While I'm not still a convert, there's enough to convince me to change, with quite a few things I had not even considered before, and surely a lot more when all that sinks in and more practice backs it up.

freeciv-web/src/main/webapp/javascript/pregame.js
```diff
@@ -1426,27 +1426,18 @@ function create_new_freeciv_user_account_request(action_type)
   var email = $("#email").val().trim();
   var captcha = $("#g-recaptcha-response").val();
 
-  var cleaned_username = username.replace(/[^a-zA-Z]/g,'');
-
   $("#username_validation_result").show();
+  var invalid_name = get_invalid_username_reason(username);
```
mchenryc: Should this be something like `invalid_reason`?

freeciv-web/src/main/webapp/javascript/pregame.js
```diff
@@ -1426,27 +1426,18 @@ function create_new_freeciv_user_account_request(action_type)
...
+  if (invalid_name != null) {
+    $("#username_validation_result").html("Your name is " + invalid_name + ".");
```
mchenryc: Pedantic semantics (:grin:): for users, maybe `("The username '" + username + "' is " + invalid_name + ".")` is slightly more clear, and gives feedback as to the name which failed. In short, it's not their username yet! Another option: `"Your choice is "...`. I favor the first.

freeciv-web/src/main/webapp/javascript/player.js
```diff
@@ -247,3 +247,27 @@ function player_has_wonder(playerno, improvement_id)
...
+  if (username == null || username.length == 0) {
+    return "empty";
+  } else if (username.length <= 2 ) {
+    return "too short";
+  } else if (username.length >= 32) {
+    return "too long";
+  }
+  username = username.toLowerCase();
+  if (username == "pbem") {
+    return "invalid";
```
mchenryc: Maybe "not available" would be more clear to a user seeing the message

freeciv-web/src/main/webapp/javascript/pregame.js
lonemadmax: Adding the name to the message is a good idea. Not a big difference here, but in hotseat, with more than one user in the same form...

mchenryc (approved):

PR #202, opened 2018-09-10
Capital locations

### LINKED ISSUE #112: Let the player find another player's (known) capital without having to middle click every visible city he owns
Various
Origin: Issue from TODO txt file. (refs #91)

Could be based on solution to Freeciv bug #23790

mchenryc:
fixes #112

### DIFF
--- a/freeciv-web/src/main/webapp/javascript/improvement.js
+++ b/freeciv-web/src/main/webapp/javascript/improvement.js
@@ -17,8 +17,6 @@
 
 ***********************************************************************/
 
-var improvements = {};
-
 // incomplete list of (well defined) building names - populate as needed
 // Remove in favor of [Effects](https://github.com/freeciv/freeciv-web/issues/208) when implemented.
 // IDs are dynamic based on ruleset
@@ -27,6 +25,26 @@ const B_PALACE_NAME = 'Palace';
 
 var B_LAST = MAX_NUM_ITEMS;
 
+const improvements = {};
+/** @private */
+const improvements_name_index = {};
+
+/**************************************************************************
+ Prepare improvements for use, resetting state from any previous ruleset
+ **************************************************************************/
+function improvements_init()
+{
+  Object.keys(improvements).forEach(k => delete improvements[k]);
+  Object.keys(improvements_name_index).forEach(k => delete improvements_name_index[k]);
+}
+
+/**************************************************************************
+ Add a new improvement definition
+ **************************************************************************/
+function improvements_add_building(improvement) {
+  improvements[improvement.id] = improvement;
+  improvements_name_index[improvement.name] = improvement.id
+}
 
 /**************************************************************************
  Returns a list containing improvements which are available from a tech.
@@ -73,14 +91,12 @@ function get_improvement_requirements(improvement_id)
 }
 
 /**************************************************************************
- Finds improvement id by exact name, returning null if not found.
+ Finds improvement id by exact name, or -1 if not found.
  **************************************************************************/
 function improvement_id_by_name(name)
 {
-  for (const id in improvements) {
-    if (improvements.hasOwnProperty(id) && name === improvements[id]['name']) {
-      return id
-    }
-  }
-  return null;
+  // 0 is a valid id, so cannot use `|| -1`
+  return improvements_name_index.hasOwnProperty(name)
+    ? improvements_name_index[name]
+    : -1;
 }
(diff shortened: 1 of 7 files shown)


freeciv-web/src/main/webapp/javascript/player.js
```diff
@@ -247,3 +247,20 @@ function player_has_wonder(playerno, improvement_id)
...
   return false;
 }
+
+/**************************************************************************
+ returns the the player's capital city, or undefined
+ **************************************************************************/
+function player_capital(player)
+{
+  return Object.values(cities)
+    .find(city => does_player_own_city(player, city) && is_capital(city));
```
lonemadmax: Can't use Object.values nor Array.find. Our friend MSIE again(*), and this time they are not translated in compilation.
This one is still readable even if more verbose in old JS. Or maybe we should start adding polyfills and forget these things once and for all.

(*) No, I haven't tested. Does one of the libraries already provide a definition for those functions?

mchenryc: Something _does_ polyfill it\*, but I'm still searching for what might be doing so.

I've reverted to the `for...in` construct in common use, leaving the polyfill question for other issues.

\* From an IE console running freeciv-web `Object.values` is defined and works as expected, while not so in a default window console.

freeciv-web/src/main/webapp/javascript/improvement.js
```diff
@@ -19,6 +19,9 @@
 
 var improvements = {};
 
+// incomplete list of (well defined) building IDs - populate as needed
+const B_AIRPORT = 0;
+const B_PALACE = 21;
```
lonemadmax: That's 23 with civ2civ3 ruleset.

lonemadmax (changes requested):

freeciv-web/src/main/webapp/javascript/improvement.js
mchenryc: Dynamic? Good catch - back to the drawing board.

FWIW, I find our PR back-and-fourth very useful in understanding "edge" (sometimes not so edge...) cases which need to be checked. Thanks again, for your help and diligence.

I expect it has come up, but is there a discussion going on anywhere around automating a test suite and how to set up a framework for testing various edge cases - something more targeted than just running an autogame for N turns?

lonemadmax: Not that I knew it depended on ruleset. I found out when looking for the source of that 21.

We had a similar problem with unit orders that depend on technologies. As a "temporary" fix, the ids were changed to a search by name (https://github.com/freeciv/freeciv-web/commit/1e3ec23ade0b6038ca7a1149381a22aa36b4b789, https://github.com/freeciv/freeciv-web/commit/60ad3e448d7eb84eeffd46ff91025a3140f1ca1d). The same search should work here with the `improvements` list.

Re test, I'm not aware of any discussion, neither automatic nor manual, not even a checklist. I guess it could be done, with a large sample of crafted savegames and casper/selenium/whatever scripts to check values and the results of different actions. It would also be a hard, thankless job.

mchenryc: I suspect the dynamic `EXTRA_*` (858cc4c6) variables I encountered in (ba71af03) are related. Any recommendations on how to get up-to speed on Effects is welcome - I see [effects.*](https://github.com/freeciv/freeciv/blob/27a777a32ef0faa0bcca819c10f2fab4dcf15d51/common/effects.c) and other refernces, but no prose description in the [HACKING](https://github.com/freeciv/freeciv/blob/master/doc/HACKING) file - I find discussions of intent and descriptions of design decisions very useful guides before diving into what I like to call "code archaeology". I'll eventually end up looking in forums or mailing lists, but if you know of discussions, it would be great to mention them in #208.

mchenryc: fixed (see commit message).

freeciv-web/src/main/webapp/javascript/player.js
```diff
@@ -271,3 +271,25 @@ function get_invalid_username_reason(username)
...
+ **************************************************************************/
+function player_capital(player)
+{
+  for (var city_id in cities) {
+    var city = cities[city_id];
+    if (does_player_own_city(player, city) && is_capital(city)) {
+      return true;
+    }
+  }
+  return false;
```
lonemadmax: You surely mean to return `city` and `undefined`.
Sorry I didn't spot this before.

lonemadmax (changes requested):

freeciv-web/src/main/webapp/javascript/player.js
mchenryc: Indeed, though you didn't miss it: this originally used `Object.values(...)`.

mchenryc: Also adjusted improvement_id_by_name to always return an int value - I realized that if a ruleset omitted a requested building, returning `undefined` or `null` could in some cases be considered the equivalent to `0` given js type-conversion.

lonemadmax (approved):

PR #207, opened 2018-09-16
Delete sourcemap files during maven initialization

### LINKED ISSUE #205: Map for minified code not generated on rebuild
As indicated in https://github.com/samaxes/minify-maven-plugin/issues/142, the source map is not generated and "linked" if it already exists when rebuilding the code. There's an easy workaround now in [build-js.sh](https://github.com/freeciv/freeciv-web/blob/develop/freeciv-web/build-js.sh#L6-L8), but not in [build.sh](https://github.com/freeciv/freeciv-web/blob/develop/freeciv-web/build.sh). I wonder if it'd be better to include the deletion somewhere in the pom instead of those scripts, or even add the clean phase to flyway:migrate and package.

mchenryc:
Due to a defect in minify-maven-plugin source maps are not regenerated
if they already exist. This pr moves the workaround from the build
script to the "official" build tool, ensuring it happens for all builds
that generate javascript

closes #205

### DIFF
--- a/freeciv-web/pom.xml
+++ b/freeciv-web/pom.xml
@@ -246,6 +246,21 @@ Rerun the sync-js-hand.js script.
 					<version>1.8</version>
 					<executions>
 						<execution>
+							<!-- workaround for https://github.com/samaxes/minify-maven-plugin/issues/142 -->
+							<id>source-mapping-fix</id>
+							<phase>initialize</phase>
+							<goals>
+								<goal>run</goal>
+							</goals>
+							<configuration>
+								<target>
+									<delete quiet="true">
+										<fileset dir="${project.build.directory}/freeciv-web/javascript" includes="**/*.map"/>
+									</delete>
+								</target>
+							</configuration>
+						</execution>
+						<execution>
 							<id>create-generated-javascript-dir</id>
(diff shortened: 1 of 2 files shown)


freeciv-web/pom.xml
```diff
@@ -245,6 +245,21 @@ Rerun the sync-js-hand.js script.
...
+							<!-- workaround for https://github.com/samaxes/minify-maven-plugin/issues/142 -->
+							<id>source-mapping-fix</id>
+							<phase>initialize</phase>
+							<goals>
+								<goal>run</goal>
+							</goals>
+							<configuration>
+								<target>
+									<delete quiet="true">
+										<fileset dir="${project.build.directory}/freeciv-web/javascript" includes="*.map"/>
```
lonemadmax: Doesn't remove `webgl/libs/webgl-client.min.js.map`. Either we wipe every map file with `**/*.map` or we specifically declare each one separately.

lonemadmax (changes requested):

freeciv-web/pom.xml
mchenryc: :man_facepalming:

lonemadmax (approved):

PR #210, opened 2018-09-25
Containerized / productized freeciv-web

doctorpangloss:
These changes should help deploy freeciv and make it easier to develop.

Configuration happens as docker environment args. There are no more `.dist` files. See the readme.

The new `docker`-based workflow supersedes local installation and the Vagrant-based development pattern. Lifecycle scripts like `start-freeciv-server.sh` have been removed. The container creation script was optimized to rebuild quickly, depending on what changes you make.

On the one hand, development iteration may be slower. On the other, testing, bug reproduction, persistence, and deployment become consistent.

The container no longer runs SSL. Use a proper SSL terminator or a similar pattern to serve HTTPS/WSS traffic.

CasperJS and its testing support have been removed, because it had not been working for a while. We'll investigate restoring this in another commit, since it works well in principle but just hasn't been caught up with the current state of the codebase.

The core installation script hasn't been seriously revised. It still installs Tomcat and Node idiosyncratically. It was modified for readability.

Logging prints to standard out, for all services, in one big stream. Tomcat, as far as I can tell, was never logging the printed stacktraces from the freeciv-web servlets correctly anyway. Logging needs to be revisited.

"Migrations" were removed. Flyway migrations are better than ever.

Hopefully these changes help us move on to solving the known issues.


andreasrosdalw:
Innovative. There are merge conflicts now. Also, why not keep or replace CasperJS? (some tests can be good)

doctorpangloss:
@andreasrosdal I can bring them back, and I can turn them back on, but then I'm really condemning this cleanup commit with a big fat test failure X!

Those automated tests are definitely a great idea, and they should be easier to run than ever. I'll restore the script and its entry in Travis after this is merged in another PR.

lmoureaux:
* Is there a migration path for installations using the old setup?
* How did you test the new setup procedure (in the README)?

Will try to find some time to give it a try, but please be patient :clock1:

doctorpangloss:
There isn’t really a migration path at the moment. The following changes will be made to better support one:

 * Publite and Metaserver servlet have to stop using MySQL as the store for information about running `freeciv-server` processes. This is the biggest obstacle right now, though this is a problem with the current codebase.
 * Save game data should be saved to a different directory than the one in Tomcat. For now, you can always bind `/var/lib/tomcat8/data` to a local directory using standard Docker run command syntax. In the future, the directory will stay fixed and be a symlink to an NFS share (supports persistence solutions like EFS).

doctorpangloss:
To test the setup procedure, install Docker for your platform, then run the commands in the `README`. The actual freeciv build takes place inside the container.

zekoz:
> * Publite and Metaserver servlet have to stop using MySQL as the store for information about running `freeciv-server` processes.

What do you propose instead?

mchenryc:
To add a bit of color, I would not approve this... _yet_, but would suggest a `dockerization` branch in this repository.

1. Since submission, I've worked to get this PR to work, but have yet to succeed (I really want to see this work!), but it requires a semi-standard (though not _universally_ standard) docker install. On my system (win7 - no apologies!) that is theoretically possible, though non-trivial enough that I am still working at it after a non-trivial amount of effort (if you have a Win7 docker install using `devicemapper` storage, ping me).
2. This is an all or nothing change from a Virtual Machine to a Docker Machine. Requiring 
Docker (or docker machine to manage a remote/VM docker service) is little different than requiring VirtualBox. However, VirtualBox _is_ a more specific start-state (raw machine vs. machine w/ some [virtual/hard]ware expectations - see point 1). I, for one, could not continue to contribute until that is worked out, either by solving the install problem, or upgrading my system (a larger ask, for any volunteer, than just contributing).
3. This image is not production ready due to several security concerns; some pre-existing, some introduced by relaxing existing behavior. This is expected, but I think we should iterate before transitioning.
4. GH tells me I've made 25 unsubmitted comments on this PR already (collected into a "prepared review"), some positive, some negative, some minor, some major, and honestly that just feels too big to drop en masse. This tells me that this PR is just too big to accept without a trial period - in a separate branch, with multiple discussions about various changes being made.

All the above are solved with a branch where this can be incorporated and iterated on. Someone will need to keep it updated with the main branch, which is the biggest risk (becoming outdated/ duplicating code in messy merges). That risk can be mitigated by separating out some of the smaller, docker-independent PRs, which provide improvements independent of the run-time environment.

doctorpangloss:
> > * Publite and Metaserver servlet have to stop using MySQL as the store for information about running `freeciv-server` processes.

> What do you propose instead?

The memory used to store which freeciv-server processes are running should have exactly the same lifecycle as the parent of the freeciv-server processes themselves. This could be as simple as asking the system for running processes.

Alternatively, the Metaserver servlet has a lifecycle almost identical to the publite lifecycle, so it should probably just be stored in ordinary Java fields in that class.

To support multiple freeciv-web servers, you'd need to use a load balancer. HAProxy has good runtime configuration support.

doctorpangloss:
> I would not approve this... yet, but would suggest a dockerization branch in this repository.

I think if this is moved to a branch, it dies.

> but it requires a semi-standard (though not universally standard) docker install.

It seems to build fine in Travis, and you'll see that the travis commands use exactly the commands in the README.

> (win7 - no apologies!) ... or upgrading my system (a larger ask, for any volunteer, than just contributing).

Gotta draw the line somewhere. On the other hand, `freeciv` is all about embracing idiosyncratic platforms. I guess that's why it's so hard to run a web service with it.

Running a working web service is a hard task!

> This image is not production ready due to several security concerns; some pre-existing, some introduced by relaxing existing behavior.

There are architectural security issues with `freeciv-web`. Has anyone fuzzed `freeciv-server`'s messaging layer?

zekoz:
I'm using hosted azure sql database for the datastore and will continue to use that for a while. If I migrate off of that, it would either be to the best free tier GCP/AWS offering or a local mysql database like play.freeciv.org and isotrident.com used.

I see no compelling reason to separate running games out of the main database.

doctorpangloss:
Let me write a more helpful plan, sorry about that:

The objective is to get `freeciv-web` running as an ordinary, scalable web service. That means there's a load balancer that sends traffic randomly to hosts hidden in a private network. When there are a lot of people playing, new hosts come up online to serve them. But from the end user's point of view, they always visit the same website.

Containers solve this objective in a way that is developer and end-user friendly. It's a solution that evolved to solve exactly the kind of engineering problems faced by complex apps like `freeciv-web`.

A host means something that serves these web pages, handles logins, and starts `freeciv-server` for you. As an end user, you should never do any IT. Zero. You don't start `freeciv-server` on your computer, you don't install anything. You visit a web page, you login, and you join a game, typically from a public listing or an invite code.

When updates are made, the **entire** host of an old version is replaced by a host of a new version. And no data is persisted (saved) inside the host. This makes things testable and reproducible, because you can totally separate the code from the data and check if you've actually fixed a bug in a particular point in time without access to a server. It helps make updates much faster, because there's no IT to test—you just run the commands in the README regardless of which point of time you're in.

The corollary of this is that `freeciv-web` has plenty of show-stopping bugs today, but reporting them is challenging. Awareness is poor when nobody is playing and everything's too hard to run. The reporter has a hard time reproducing. The repliers have a hard time reproducing.

To accommodate the separation of data and code engineering-wise, no data can ever be saved inside the host. It's either saved in a database or a networked filesystem. Concretely, `freeciv-server` needs to save games and score logs frequently to a shared filesystem. The metaserver data needs to reflect which `freeciv-server` processes are actually running, which means the hosts have to be aware of each other. The best way to do that in Java is to have a cluster-wide event bus using something like Hazelcast or Atomix. Whatever is running the `freeciv-server` process responds to queries for free processes to run games, and it stops answering when it's shut down.

There needs to be some kind of routing, so that the client can initiate a web socket connection to `ws://freecivweb.org/games/839121839` and the dumb load balancer doesn't need to be aware of what `839121839` means (none of the cloud service ones support logic like that). You can write a load balancer, or just configure the host that receives the connection randomly to route the traffic to the right host. That means you pass websocket traffic, sometimes, over your internal event bus. No big deal.

freeciv-web/src/main/java/org/freeciv/servlet/CivclientLauncher.java
```diff
@@ -28,51 +30,53 @@
...
-				break;
-			case "pbem":
-				gameType = "pbem";
-				break;
-			default:
-				response.setHeader("result", "invalid port validation");
-				response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
-						"Unable to find a valid Freeciv server to play on. Please try again later.");
-				return;
+				case "new":
```
zekoz, line 62: Please revert the whitespace only changes.

zekoz:
I have commit access to this repo now. This PR is +770 −1,673 loc. To be able to merge this, I want this broken up into multiple PRs, there is just too much here for me to review at once. Some comments:

- This should be more than two containers. You can make one container as a first PR just be fcwebfe (the java web app). I have been playing around on my own with a Dockerfile for civserver (freeciv server binary). Everything else should go into a new container. Some of these free-running python processes like earth and pbem should be merged into fcwebfe.

- Leave out the whitespace-only changes.

- It looked like you removed a lot of stuff from the readme that is still relevant.

- I don't like the properties changed to getenv- That makes it much harder for me to run in azure. I will continue to run the fcwebfe as a hosted service, either azure web app or google appengine for the foreseeable future. Docker is great for local testing, but the production environment needs to remain deployable.

PR #211, opened 2018-09-30
Provisional game settings for game XVII

qrtv2:
63% land and fully random height generator gives big land areas, will also let navies have a key role. 

I tried experimenting with pseudo-fractal (50% land) but unless temperate map is off, it will mess up the map badly. Temperate map off would mean that many players end up in the middle of the dessert. 

I set revolution length to 1 and space race off.


qrtv2:
Sent to wrong repository.

Issue #177, opened 2018-08-06
Issues reported by lgtm.com code review


andreasrosdalw:
https://lgtm.com/projects/g/freeciv/freeciv-web/

https://lgtm.com/projects/g/freeciv/freeciv-web/alerts/?mode=list

Some of these can be good tasks for new developers interested in small, fun programming tasks.

lonemadmax:
Don't be overwhelmed by the quantity. Just take a stab at an area you feel confident with or of which you want to know more. No need to solve them all at a time!

devansh289:
Hey, I'm a beginner and would love to work on these.

ErakordnePraht:
Can anyone tell me what i did was correct? I'm not very familiar with python and i don't know how to test my integration.

zekoz:
@truberton @devansh289 @mchenryc Do you prefer IRC or discord for talking to the other developers?

mchenryc:
@truberton you should submit a PR for your change, which will instigate the discussion of it.

If you're not familiar with the fork-pull model used by freeciv-web, github has lots of documentation, start with: https://help.github.com/articles/creating-a-pull-request-from-a-fork/ . Submit a PR, and we can discuss forks/branches, submitting PRs and referencing issues there.

fxedel:
To mention @mchenryc here:

> PR should generally have a purpose other than"fixing code" - the thing with changing lines just for the sake of style or syntax changes is that it hides the previous "logic" change. In other words, the why a line of codes existence, can usually be determined by the comments in the previous commit for that line and those around it - git blame is used to see why a change was made, understand the context, and determine if, and how it can/should be modified. To change code simply for the sake of whitespace, or re-arranging lines is frowned upon, because it unnecessarily hides/buries history.

> That said, personally I love deleting dead code and encourage it when possible! As long as it's actually dead code, deletion doesn't change the history of the working code around it.

> The exception to "don't change for the sake of" is when explicit cleanup tasks are warranted/desired. There is an open issue to clean up code smells with many commits already accepted towards that end, and others actively being worked on. Such changes are welcome - reference the issue in your PR, and keep it on task.

Maybe one could add this to the initial comment.

GaeaKat:
https://lgtm.com/projects/g/freeciv/freeciv-web/alerts/?mode=list is forbidden when you try to look
Going from https://lgtm.com/projects/g/freeciv/freeciv-web/ seems to work better

PR #214, opened 2018-10-05
removed redundant import

ErakordnePraht:
see #177

### DIFF
--- a/scripts/freeciv-img-extract/img-extract.py
+++ b/scripts/freeciv-img-extract/img-extract.py
@@ -14,7 +14,7 @@
 '''
 
 from os import environ, path
-os.environ['PYTHON_EGG_CACHE'] = '/tmp'
+environ['PYTHON_EGG_CACHE'] = '/tmp'
 
 import argparse
 import configparser
--- a/scripts/freeciv-img-extract/img-extract.py
+++ b/scripts/freeciv-img-extract/img-extract.py
@@ -13,14 +13,13 @@
    GNU General Public License for more details.
 '''
 
-import os, sys
+from os import environ, path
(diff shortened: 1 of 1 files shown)


zekoz:
I prefer import os at the top and using os.environ below for readability.

mchenryc:
Since `os.path` is a module, you're able to simply:

    import os.path;
    os.environ['PYTHON_EGG_CACHE'] = '/tmp'

...which imports `os` as well, because "[`os.path` is funny](https://stackoverflow.com/a/2725195/80714)." Following the "explicit is better" mantra though, import both modules (os, and os.path). While `import os, os.path;` would work, one import per line is preferred (with which I agree), so I would probably go with:

    import os;
    import os.path;
    os.environ['PYTHON_EGG_CACHE'] = '/tmp'

I agree w/ zekoz about qualifying `os.environ...`. The trend is to be more explicit, so instead of `import x from y;  x(...);`, use qualified versions of properties such as `os.environ`.

It's rather inconsistent, even in this codebase - which if you _made_ consistent, would be great. This pr could simply make importing os and os.path consistent.

As to testing - have you built the Vagrant Box? There is no explicit test for this script, but it is used to "import" `freeciv` files into `freeciv-web`, so rebuilding your vagrant box would test it. Better yet, run it directly (but you'll probably want to have built the Vagrant box anyway). The build runs it like this (though for testing, the output dir is not relevant).

    cd scripts/freeciv-image-extract
    img-extract.py -f ../../freeciv/freeciv -o ../../freeciv-web/src/derived/webapp

andreasrosdalw:
Thanks

PR #219, opened 2018-11-03
Fix: use extras list to show resources

### LINKED ISSUE #41: Special resources are not shown correctly from a loaded savegame in Freeciv-web
Special resources are not shown correctly from a loaded savegame in Freeciv-web.

http://forum.freeciv.org/f/viewtopic.php?f=24&t=3497

lonemadmax:
The resource field only makes sense in the server, it is used to know
a tile has a resource even if terrain changes and it's not valid, in
case it changes back. It's not even rebuilt for the private map when
loading a game. Real valid resources are in the extras list.

Fixes #41 (resources not shown from a loaded savegame)
Fixes fish on swamp and other such resources in changed terrain


lonemadmax:
Works for me, but the webgl change is untested.

PR #224, opened 2018-11-25
Disable serial port, avoiding Win10 issues

mchenryc:
Vagrant creates a log file for the serial port, which, on Windows 10 may
not be able to be made. This is likely related to use of WSL, and root
file mapping.

As freeciv has no need for the serial port, disabling it solves the
issue.


lonemadmax:
Ref https://github.com/hashicorp/vagrant/issues/8604
This removes the boot log file that might help if the vm doesn't boot, but I guess if anyone needs it they can remove change.

PR #227, opened 2018-12-04
Update Travis to Ubuntu Xenial

mchenryc:
Builds began failing tonight, due to issues updating Oracle JDK 8 during Travis build.

This patch updates the Travis build to Xenial ([recently released](https://blog.travis-ci.com/2018-11-08-xenial-release) on Travis), which has updated versions that are not failing.

Given the Vagrant box uses Bionic (2018), upgrading from Trusty (2014) to Xenial (2016) on the CI server should not present an issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: xenial
 
 language: c
 


lonemadmax:
Happened to me last week and didn't pay attention because it went right the next day and it was clearly a hiccup.

Now I see it's *Oracle's* JDK. The one we install and for which add a ppa if necessary is OpenJDK. But Travis includes webupd8's ppa and has OracleJDK already provisioned. When we run the update it tries to also update that package and that's where it fails. I guess we could remove it just as we do for postgresql, mysql and its data, and stay with Trusty, but I don't think we have a reason for that. Especially with Trusty going EOL in four or five months.

play.freeciv.org run on Trusty, and I think Andreas had scripts to check a commit result before updating the server. If anyone is running freeciv-web on Trusty today, please shout. No, don't shout, just upgrade.

I don't think there's a difference in having one or another build OS as long as we don't touch the installation part. And when we do, we should test it with all the systems we consider "supported", not just one. FWIW, I'm running Debian.

Re the Vagrant box, I think Andreas used to update to new releases. At least the history shows it has gone through all versions since Saucy. I've slacked here. I'm gonna make up an excuse and say it's just LTS releases now.  Or at least until it is updated to a non-LTS one :-P

mchenryc:
> No, don't shout, just upgrade.

😁 From what I gather, the Oracle thing is a recurring issue. Appears, annoys, disappears. I rather like the idea of _removing_ oraclejdk from the travis image to avoid the annoyance.

> ... say it's just LTS ...

👍  - really, there's no reason to ride the bleeding edge here.

lonemadmax:
> I rather like the idea of removing oraclejdk from the travis image to avoid the annoyance.

I tried. Removing it complained about the value of JAVA_HOME. Setting JAVA_HOME complained about the value of JAVA_HOME. I didn't put more effort into finding out what I was doing wrong and what other configuration bits had to be changed, I could put up with waiting a day and firing a rebuild.

It seems the only external repositories for the moment in Travis' Xenial are postgresql and git-lfs. The included java version is openjdk11.

Oraclejdk8 was not just "not openjdk". It was distributed through an installer in a package provided by a third (fourth?) party that maybe had playing catch-up as their best option.

PR #231, opened 2018-12-24
Stricter auth

lonemadmax:
Correct authentication checks in the server, or at least make them less easy to bypass.


mchenryc:
@lonemadmax Just pulled this locally, and I am no longer able to use an ad-hoc username - in other words, this appears to force going through the registration process to play run/play locally.

Was that intended? For my sake, I rather liked the ability to simply start a singleplayer game with nothing more than entering a name.

mchenryc:
Previous comment of issue deleted - I rebuilt everything from scratch and it appears to work fine. My issue was likely some changes to, or for freeciv updates not being reverted correctly.

PR #232, opened 2019-01-03
Cleanup status script

mchenryc:
While working updating to freeciv server, I found these changes useful.

Overall status will appear as:
```
vagrant@ubuntu-bionic:/vagrant/freeciv-web$ ../scripts/status-freeciv-web.sh
Checking that Freeciv-web is running correctly... (-v for verbose)

Checking nginx                                     OK!
Checking Tomcat                                    OK!
Checking freeciv-web on Tomcat                     OK!
Checking freeciv-proxy directly                    OK!
Checking freeciv-proxy through nginx               OK!
Checking WebSocket directly to Tornado             OK!
Checking WebSocket through nginx                   OK!
Checking freeciv-web / publite2                    OK!
Checking webclient.min.js generation               OK!
Checking tileset generation                        OK!
Checking metaserver on PHP server                  OK!

--------------------------------
Check of Freeciv-web is done!
```

### DIFF
--- a/freeciv/dl_freeciv_default.sh
+++ b/freeciv/dl_freeciv_default.sh
@@ -1,16 +1,20 @@
-#!/bin/sh
+#!/bin/bash
 
 # Places the specified revision of Freeciv in freeciv/freeciv/
 # This is the default. The script dl_freeciv.sh will run instead of
 # dl_freeciv_default.sh if it exists.
 # If you want to modify this file copy it to dl_freeciv.sh and edit it.
 
+echo "Updating freeciv to commit $1"
+
 # Remove old version
+echo "  removing existing source"
 rm -Rf freeciv
 
 # Download the wanted Freeciv revision from GitHub unless it is here already.
 # The download step saves having to merge in Freeciv's history each time the
 # Freeciv server revision is updated.
+echo "  fetching missing revisions"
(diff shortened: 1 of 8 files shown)


lonemadmax:
In status-freeciv-web.sh:
- Please review the use of `logfile`: in `checkURL` and `checkWS` it is read as a parameter but not used, whereas in `checkWebStatus` it is used without assignment.
- The webclient, tileset and metaserver checks lost the `--head` parameter to `curl`. Was that on purpose?
- If any of those are not right and you are changing this again, could you also remove the reference to PHP? It's been ages since the metaserver was rewritten.

mchenryc:
* logfile: thanks - forgot to remove logfile as a parameter to those functions after the last refactoring which changed logging so all tests (urls, websockets, pids, and services) use stderr/out in a common way from within `checkStatus` (curl sends info to stdout, when `-o logfile` is not used).
* added `--head` param back - using it saves bandwidth, though curl gives the same exit status regardless
* Ha! Yeah, actually I just removed the metaserver test entirely, given it's just another servlet in Tomcat. It gives us no more info about services being up/down than the "freeciv-web on Tomcat" test - LMK if you feel like it needs to be there

I also added local variable naming to a couple more functions, and made parameter order consistent, to make them easier to follow.

Lastly, a couple more start/stop scripts were changed to specify bash as well - see added commit.

PR #233, opened 2019-01-07
My First Commit. Only javadoc and tabs (using Eclipse)

QuiqueOlaso:
This is one of my first projects in GitHub and the first one sharing with other people, so i thought to start with something easy ....

### DIFF
--- a/nginx/sites-available/freeciv-web
+++ b/nginx/sites-available/freeciv-web
@@ -137,6 +137,9 @@
           proxy_redirect     off;
 	}
 
+        location = /meta/metaserver {
+          deny all;
+        }
 
 	# proxy for app-server on port 8080
 	location ~ / {


QuiqueOlaso:
Also some refactoring ... sorry i forgot!

lonemadmax:
Thank you for thinking about freeciv-web. I hope the slow pace doesn't discourage you to continue.

In the formal aspect, valid not only for this PR and repo:
- Nobody is interested in whether this is your first commit or the tools you use when they browse the changes. I guess it's OK to mention it in the discussion, so that you get more detailed explanations and some hand holding, but keep the title short and to the point, just telling what the patch is about.
- Make different changes in different commits. There's some variation on what different projects want and accept, but as a general rule each commit should have just one purpose. As I see it, this should be at least one commit for the javadoc and another one for the constants.
- Use branches. When you are making changes, keep the main branch alone and make a new one with the changes you want. It makes your life easier if they need reworking or your PR is not accepted. It also makes other people's life easier if they have cloned your repo.

Now to the contents.

I do like documention that explain what the different parts are and how they work. But stuff like this:
```diff
+ // TODO: Auto-generated Javadoc
+ /**
+  * The Class Game.
+  */
  public class Game {

+ 	/** The host. */
  	private String host;

+ 	/** The port. */
  	private int port;

[...]
+ 	/** The Constant gameTypes. */
  	private final static Map<String, Integer> gameTypes = new HashMap<>();
[...]
+ 	/** The p. */
  	private Pattern p = Pattern.compile(PATTERN_VALIDATE_ALPHA_NUMERIC);

+ 	/** The Constant mapDstImgPaths. */
  	private static final String mapDstImgPaths = "/var/lib/tomcat8/webapps/data/mapimgs/";
```
is noise to me, with this other kind even being wrong:
```diff
  /**
   * Displays the multiplayer games
-  *
-  * URL: /meta/metaserver
+  * 
+  * URL: /meta/metaserver.
   */
```

The constants extraction seems a good thing. I wonder if maybe it should go the extra mile to a getDBConnection factorization.

I'll wait for others with more experience in good practices to say something, though. It wouldn't be the first time that I'm proven wrong in what I think is the right way of doing things.

mchenryc:
I agree with everything mentioned by @lonemadmax.

Especially the beginning: **don't let critique of your initial PR discourage you.** Every project has it's own style of working - though individuals may prefer different styles. As a new submitter, your job is to learn the styles for this project, and _why_ they're in place. As you learn, your submissions are accepted, and your work becomes trusted by other members, you can make arguments for changing the existing rules. I consider myself transitioning into the second stage.

See [CONTRIBUTING.md](https://github.com/freeciv/freeciv-web/blob/develop/CONTRIBUTING.md).

The easiest way to become a productive member, and understand a project, is to submit **small** commits making meaningful changes, in the current style of the project.

Primary takeaways for this PR:
* Use meaningful commit messages.
* Don't "fix" style problems without purpose - this buries the most recent commit message, which is more important to have available than fixing style issues.
* Small PRs can be reviewed and accepted _much_ faster than large PRs.
* Don't add useless comments

The end result is we're going to ask you to drop _most_ of this PR (not all). The rest of this comment is the "why".

The real change here is "Use constants for JDBC connections" which will be maybe 50 lines or so, in a single commit. Drop everything else (see below).

----

**Git commit messages**: I use the 50/72 rule for commits, and encourage others to as well. Read these:
* https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53
* https://chris.beams.io/posts/git-commit/ - Long, but explains why many projects have seemingly "odd" rules about commit messages
* https://github.com/freeciv/freeciv/blob/master/doc/CodingStyle Our parent project's style guide. Worth being familiar with, though _many_ points in it simply don't apply to Java and Javascript. Much of the JS here was ported _directly_ from C code in FC with only minor syntactic changes, so you'll see patterns that don't necessarily make sense in FCW.

**Commits**: Multiple commits in a single PR are great, and multiple PRs are often better. Everything in a PR should have a common purpose/theme. A big advantage of submitting multiple smaller PRs is that a problem in one won't prevent the other PRs from being merged.

The point is to keep the _most recent commit_ message meaningful for any given line. `git blame` is a great way to understand why code is there, but it only shows the _most recent_ commit. With meaningful commits this is often all that is needed (otherwise one needs to start digging through the full log. Some editors make this information easily available - but not everyone uses those so be kind to the minimalists.

**Whitespace changes**: Avoid. If you're making actual logic/behavioral changes, fix the lines you must change anyway. Fixing an adjacent line amidst others can be acceptable, but usually, leave them until later. Don't bury meaningful commits without good cause.

_Tabs vs. spaces_. [I hate tabs](http://tarantsov.com/hackers-coding-style-guide/why-tabs-should-be-avoided.html) - screw the character in "Silicon Valley". Yet... not all agree with me \*. Someday I want a commit on this project with the sole purpose of changing all tabs to spaces, and forever banning tab indentation. Until then - don't change it just for the sake of changing it. Turns out much the Java code here uses tabs, but some doesn't. Keep consistent within a file until such time as we can "fix" that 😉. My editor is smart enough to follow the current file's convention, but I also have the "show whitespace" feature enabled so I can see what is being used.

\* My take is that the java code started w/ tabs, but is (slowly) moving toward all spaces - at least that's my take on the current inconsistency. If a task changes the vast majority of a file, it may make sense to convert that file. In that case, 2 commits may be worthwhile - first the whitespace only change, then the logic changes. This keeps the last commit the meaningful logic changes, and marks the few other lines as "meaningless... you need to look at the previous change".

**Code Comments**: become outdated, misleading, and ignored - less is more, especially as this is an application, not a library consumed by other applications (no need to publish a reference API). Comments _can_ provide valuable understanding of why a function exists, and how it works in relation to other parts of the ode. But comments that reiterate names are worse than worthless detrimental.
1. keep functions short
2. name functions so the behavior is obvious.
3. only use comments to explain the non-obvious

----

Final word - Personally, I totally empathize with wanting to "clean things up" and make lots of little fixes to make code consistent or follow the "right style". I avoid the impulse though, as I will spend as much time removing those changes, so the meaningful work is not buried in unrelated noise. Feedback/review is critical, which can't happen in large commits full of unrelated, non-strictly-necessary changes.

PR #234, opened 2019-01-21
Fixed Docker build issues

Adriankhl:
In Dockerfile, I have added back the missing config folder.

Also, running ./start-freeciv-web.sh with Docker gives some locate errors, I have modified the Dockerfile to fix the locate issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,12 +21,24 @@ ADD tests /docker/tests
 ADD music /docker/music
 ADD blender /docker/blender
 ADD nginx /docker/nginx
+ADD config /docker/config
 
 ## Install relevant packages
 
-RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get -y upgrade && apt-get install -y sudo lsb-release
+RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get -y upgrade && apt-get install -y \
+    sudo \
+    lsb-release \
+    locales
 
 EXPOSE 80 8080 4002 6000 6001 6002 7000 7001 7002
+
+## Fix locate issues
+
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
+    dpkg-reconfigure --frontend=noninteractive locales && \
+    update-locale LANG=en_US.UTF-8
+
+ENV LANG en_US.UTF-8 
 
 ## Create user and ensure no passwd questions during scripts
 


Dockerfile
```diff
@@ -21,13 +21,24 @@ ADD tests /docker/tests
...
 
 ## Install relevant packages
 
 RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get -y upgrade && apt-get install -y sudo lsb-release
 
 EXPOSE 80 8080 4002 6000 6001 6002 7000 7001 7002
 
+## Fix locate issues
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
```
mchenryc: instead of adding another apt-get install (which does another update) just add 'locales' to the install instr. above

Then, as long as you're editing that line, put each package installed on it's own line (adding trailing `\` as necessary of course) - this makes adding/removing packages easier to read/review.

mchenryc (changes requested):

Dockerfile
Adriankhl: Done, thanks for the comment.

Dockerfile
```diff
@@ -21,13 +21,24 @@ ADD tests /docker/tests
 ADD music /docker/music
 ADD blender /docker/blender
 ADD nginx /docker/nginx
+ADD config /docker/config
 
 ## Install relevant packages
 
-RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get -y upgrade && apt-get install -y sudo lsb-release
+RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get -y upgrade && apt-get install -y sudo \
```
mchenryc: Personally, I'd go with 
```
...install \
    sudo \
    ...
```
as it keeps the indent standard, and _every_ dep is consistently on it's own line. Minor, and if it's merged without that, no biggie.

mchenryc: Also - I would recommend [squashing](https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git) these commits.

This is a nice, very discrete PR, with only one aspect of change (as preferred!), and no need to keep PR corrections in the project changelog.

Adriankhl: My first time to use squashing, seems like it is working.

mchenryc: Just to clarify my earlier comment, standard indent is ~4 spaces~ 2 or 4 spaces (depending on the file's language), so the suggestion was to keep the continuation lines (dependencies in this case) at that, even though the line is long. I ought to have put more text in my example, like so:
```
RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get -y upgrade && apt-get install -y \
    sudo \
    lsb-release \
    ...
```

The idea is not particular to Dockerfiles - it's that the long line is... important, but less likely to change than individual additions to/removals from the list below. Consistent indentation w/ the codebase keeps lists like this consistent and legible throughout the codebase, and changes are easy to understand in diffs.

Again, it's a bit trivial so not a big deal if @lonemadmax merges without that change, though if you get a chance, feel free to practice your squash/force-pushing again 😁.

Good PR, thanks!

Edit: "fixed" spacing advice

PR #237, opened 2019-02-10
Separate longturn savegames

lonemadmax:
Make a new directory lt/game_name so that longturn type savegames don't get all mixed in the root data savegames directory. This way they are easier to track and to reload in case of server restart.

### DIFF
--- a/publite2/init-freeciv-web.sh
+++ b/publite2/init-freeciv-web.sh
@@ -13,19 +13,22 @@ if [ $5 = "pbem" ]; then
    pbemcmd="--Ranklog /var/lib/tomcat8/webapps/data/ranklogs/rank_${2}.log "
 fi
 
+savesdir=${1}
 quitidle=" -q 20"
 if [[ $6 == *"longturn"* ]]; then
   quitidle=" "
+  savesdir="${savesdir}/lt/${6}"
+  mkdir -p "${savesdir}"
 fi
 
-export FREECIV_SAVE_PATH=${1};
+export FREECIV_SAVE_PATH=${savesdir};
 rm -f /var/lib/tomcat8/webapps/data/scorelogs/score-${2}.log; 
 
 python3 ../freeciv-proxy/freeciv-proxy.py ${3} > ../logs/freeciv-proxy-${3}.log 2>&1 &
 proxy_pid=$! && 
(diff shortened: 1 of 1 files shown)


publite2/init-freeciv-web.sh
```diff
@@ -13,19 +13,22 @@ if [ $5 = "pbem" ]; then
...
+savesdir=${1}
 quitidle=" -q 20"
 if [[ $6 == *"longturn"* ]]; then
   quitidle=" "
+  savesdir="${savesdir}/lt/${6}"
+  mkdir -p "${savesdir}"
 fi
 
-export FREECIV_SAVE_PATH=${1};
+export FREECIV_SAVE_PATH=${savesdir};
```
mchenryc, line 24: Can you confirm this env variable is still necessary? We use `--saves` below, but I _believe_ this is the directory init scripts are loaded from (e.g. `--read` param below), which is not overridden by `--saves`.

If the above is correct what do you think of removing the env variable anyway, and specifying the `--read` parameter absolutely (e.g. "${savesdir}/pubscript_${6}.serv"), so all arguments are passed in the same manner, via cli?

lonemadmax: From what I understand in the freeciv-server manual page, the `--saves` option tell the server where to place savagames, while the envvar gives it a list of directories from where to load savegames. Nothing to do with `--read`.

And that means I should probably add the base dir, in case there's a longturn game that loads a scenario. Or we can leave it as is, and expect the game master to put the scenario in its lt directory.

The `pubscript*.serv` scripts are all in the `publite2` directory from which init-freeciv-web is launched. You do have a point in that the parameter for the script name is bizarre (for `${HOME}/freeciv-web/publite2/pubscript_singleplayer.serv` the arg is singleplayer). Changing that would mean changing publite2, which I'm not against, but I'd rather not do here.

mchenryc: In that case, just leave it as is.

I have a branch to play with the test harness, which triggers `init-freeciv-web.sh` for individual test instances (rather than rely on the publite2 polling metaserver). The Publite2 changes are motly cleanup, with the notable change of propagating the freeciv stdout, and using expect to await the server's ready state. This allows confirmed game readytime, without arbitrary timeouts which just "hope" the game is ready.

Once you're current patches are in, I'll push a PR for that work, which should be good as a standalone patch.

PR #238, opened 2019-02-10
Explicit parameters separation in civlauncher

lonemadmax:
Give `subprocess.call` an array, instead of a string with unescaped arguments for the shell to (mis)interpret.


mchenryc:
LGTM

PR #242, opened 2019-02-14
Improve Dockerfile and add docker-compose.yml

AlwinGarside:
Made some improvements to the Docker stuff.

  - Made improvements to the Dockerfile in line with the [best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) and personal experience.
      - Changed `ADD` to `COPY` because that is more concise.
      - Removed `apt-get upgrade` because it's not recommended.
      - Added removal of cache files to reduce image size.
      - Put all the generic and slower commands before the quick commands to prevent that everything is rerun every time a file changes.
  - Keep `docker` user in `sudoers` to prevent it asking for password when starting the start script.
  - Added `docker-entrypoint.sh` which automatically starts freeciv when starting the container.
  - Created `docker-compose.yml` to easily start container with docker-compose.
  - Added `.dockerignore` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-﻿THE FREECIV-WEB PROJECT
+THE FREECIV-WEB PROJECT
 -----------------------
 
 [![Build Status](https://api.travis-ci.org/freeciv/freeciv-web.png)](https://travis-ci.org/freeciv/freeciv-web)
@@ -164,27 +164,25 @@ Start and stop Freeciv-web with the following commands:
 
 All software components in Freeciv-web will log to the /logs sub-directory of the Freeciv-web installation.
 
-### Freeciv-Web alternative Docker image
 
-1. Build local dockerfile based on debian - this will take a significant amount of time (10 - 20 minutes)
+### Running Freeciv-web on Docker
 
-docker build . -t freeciv-web
+Freeciv-web can easily be built and run from Docker using `docker-compose`.
 
-2. Run docker image including required ports on your host
+ 1. Make sure you have both [Docker](https://www.docker.com/get-started) and [Docker Compose](https://docs.docker.com/compose/install/) installed.
 
-docker run -i -t --user docker -p 8888:8080 -p 80:80 -p 7000:7000 -p 7001:7001 -p 7002:7002 -p 6000:6000 -p 6001:6001 -p 6002:6002 freeciv-web
+ 2. Run the following from the freeciv-web directory:
 
-3. Start Freeciv-web with:
+    ```sh
+    docker-compose up -d
+    ```
 
-./start-freeciv-web.sh
-
-Answer prompt for docker sudo password with "docker"
-
-4. Connect to docker via host machine using standard browser
+ 3. Connect to docker via host machine using standard browser
 
 http://localhost/
 
 Enjoy. The overall dockerfile and required changes to scripts needs some further improvements.
+
 
 Freeciv-Web continuous integration on Travis CI 
 -----------------------------------------------
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,12 @@
+.dockerignore
+.gitattributes
+.gitignore
+.mailmap
+.travis.yml
+ADVANCED.md
+CONTRIBUTING.md
+Dockerfile
+README.md
+README-One-Turn-Per-Day.md
+TODO
+Vagrantfile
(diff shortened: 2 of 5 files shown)


lonemadmax:
I can say it builds and runs, but I'd rather someone else who knows docker chimes in.

I think it needs docs updating. At least step 3 (start freeciv-web with `./start-freeciv-web.sh`) is not needed anymore.

And I didn't get the savegames in a second run, which is probably my fault, as I don't know how that is supposed to work.

AlwinGarside:
I have a lot of experience with Docker, but I can understand you’d like a second opinion.

I’ll update the docs to reflect the changes made.

As for the savegames, I’ll double check.

lonemadmax:
Oh, well. My fault, as expected. I did the build and run dance, but it seems we need docker-compose now, as should have been clear after reading "Created `docker-compose.yml` to easily start container with docker-compose."

If no one shouts, I'll merge this once the docs are updated.

I also appoint you the docker gatekeeper, may you accept the position. Keep in mind it is currently treated as a second-class citizen. I have more than once introduced changes in other areas and not noticed I had broken the docker build until much later, when someone else sent a correction.

AlwinGarside:
Just remembered I had updated the docs, but hadn't committed / pushed it yet. :-)

I don't mind being the Docker gatekeeper. I'll try and keep an eye on changes that may break Docker.

On a related note: I noticed there seems to be an official freeciv/freeciv-web repository here:
https://hub.docker.com/r/freeciv/freeciv-web

However there are currently no images there. Might be a nice idea to set it up to automatically build the Dockerfile there.

README.md
```diff
@@ -164,28 +164,26 @@ Start and stop Freeciv-web with the following commands:
 
 All software components in Freeciv-web will log to the /logs sub-directory of the Freeciv-web installation.
 
-### Freeciv-Web alternative Docker image
 
-1. Build local dockerfile based on debian - this will take a significant amount of time (10 - 20 minutes)
+### Running Freeciv-web on Docker
 
-docker build . -t freeciv-web
+Freeciv-web kan easily be built and run from Docker using `docker-compose`.
```
lonemadmax: s/kan/can/

lonemadmax:
I don't know who did the hub.docker thing.

I'm not sure freeciv-web is really ready for distribution as a docker container. As it is now, it'd be a multigigabyte lump with OS updates, development tools, git history and all of freeciv-web. Not really containerized.

Others tried to move in that direction before, like in https://github.com/freeciv/freeciv-web/pull/210, but nothing clear came out of it.

Feel free to add it to any build or distribution service, though.

AlwinGarside:
Ah yes, you're right. I built my own image and it weighs in at 2GB:
https://hub.docker.com/r/yogarine/freeciv-web/tags

The solution is to do a multi-stage build where the development environment is set up and freeciv-web is built in a temporary image, and then in the final build only the required libraries and the freeciv-web binaries are installed.

To do this I'll need to figure out exactly what the run dependencies are (as opposed to the build dependencies) and which files need to be copied over after the build is done. Any help with that would be welcome.

lonemadmax:
Sometimes I wish I was more verbose. Could you squash the typo fix into the previous commit? No need to have that in history.

AlwinGarside:
> Could you squash the typo fix into the previous commit? No need to have that in history.

Squashed it!

lonemadmax:
> To do this I'll need to figure out exactly what the run dependencies are (as opposed to the build dependencies) and which files need to be copied over after the build is done.

I *think* the runtime would need:
- mysql, nginx, tomcat, python3, curl, basic OS stuff and their dependencies and configuration files.
- python packages: tornado, mysql-connector-python and their dependencies. I have some pending changes to install those through pip to the user site, so they should be available (with quite a few more used when rebuilding freeciv) in $HOME/.local/lib/python3*
- freeciv: it is built into $HOME/freeciv
- freeciv-web webapp: I guess /var/lib/tomcat8/webapps/freeciv-web.war would be all.
- From the repo: LICENSE.txt, freeciv-earth, freeciv-proxy, logs (just the directory to write them, it can be mounted from out of the image), pbem (there's the scripts and one changing data file: pbem-games.json, maybe we should move it somewhere else?), publite2 and part of scripts. Certainly not blender (116M), freeciv (402M in my case, I think I have all the history), music (31M), freeciv-web (205M in my case) and .git (728M in my case).

Someone else will have to say what java packages are needed. Probably maven takes care of installing them when building the app.

If the source used to build it has to be in the docker image, then add at least freeciv-web/src and freeciv/freeciv.

PR #244, opened 2019-02-15
tomcat: accept utf8 as default

lonemadmax:
This fixes a problem with non-US-ASCII characters in player and nation names reported from freeciv to the metaserver. It's been in real use for several weeks, but I don't known if it's The Right Way.
- Maybe it breaks setups running freeciv with a different encoding.
- Maybe freeciv should be changed to tell the encoding it is using when sending data to the metaserver.
- Maybe the filter should be applied only to the metaserver URL.
- Maybe...


andreasrosdalw:
it solves the problem...

PR #246, opened 2019-02-24
Python dependencies

lonemadmax:
Make dependencies explicit with requirements files, use pip to install them and update both mysql-connector and tornado.


mchenryc:
LGTM

PR #247, opened 2019-02-25
fix: activate map key handler in map tab

### LINKED ISSUE #245: Keyboard shortcuts not working
The keyboard shortcuts are not working, for example pressing G for goto.

http://144.76.115.248

lonemadmax:
#223 separated different key handlers so that pressing a key in a dialog
fired only the action for that dialog if defined, and didn't send orders
to units or fired other events.

Sometimes the focus doesn't return to the map element after the dialog
is closed, making only the global handler active, and leaving the user
with a broken keyboard interface until the map is clicked or gets the
focus for some other reason.

We now check in the global handler whether the map tab is active and no
dialog is on sight, and run the map kay handler in that case.

fix #245

### DIFF
--- a/freeciv-web/src/main/webapp/javascript/control.js
+++ b/freeciv-web/src/main/webapp/javascript/control.js
@@ -256,9 +256,6 @@ function control_init()
   /* prevents keyboard input from changing tabs. */
   $('#tabs>ul>li').off('keydown');
   $('#tabs>div').off('keydown');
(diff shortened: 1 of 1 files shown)


lonemadmax:
I still think this is a hack and proper "window" stacking and focus handling is needed, but that's a bigger endeavor and this seems to solve the problem at hand.

PR #249, opened 2019-02-26
Longturn autoreload

lonemadmax:
Automatically load the last savegame when a longturn game is restarted.
To use it, add a comment with "autoreload" in the start script.
Whether we should use the modification time (as in the current patch), lexicographical order or anything else is open to discussion.

### DIFF
--- a/publite2/init-freeciv-web.sh
+++ b/publite2/init-freeciv-web.sh
@@ -36,6 +36,14 @@ savesdir=${1}
 if [ "$5" = "longturn" ]; then
   savesdir="${savesdir}/lt/${6}"
   mkdir -p "${savesdir}"
+
+  grep -q '^#\s*autoreload\s*$' "pubscript_${6}.serv"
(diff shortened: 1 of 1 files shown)


andreasrosdalw:
This is awesome.

I wish this also made the game automatically start.

Lexxie9952:
Yes if game automatically restarted, that would be even more awesome. But this is still so great, for not having to dig to find savegame file !  Thank you so much !

andreasrosdalw:
Looks good, can you merge this?

lonemadmax:
> I wish this also made the game automatically start.

This might be dangerous. The game is restarted without notice and what was done in that turn is lost. You may also have to edit the savegame to solve whatever problem there was, and have a restart-die loop meanwhile.

But you can have it. Just put `start` at the end of the serv file. OK, it's freeciv itself and not this commit the one that makes the game start, but it's there :-)

PR #265, opened 2019-03-25
Pick the nation in ongoing longturn games

RhueOfL:
Resubmitting Pull Request 261
with new features -
Only the connections taking over "New Available Players" are allowed to pick their nation.
Button "Pick a random nation" now actually draws a random nation in order to avoid a possible
doubling of a nation in-game.


andreasrosdalw:
Thanks! Perfect.

PR #267, opened 2019-03-26
Refactoring queries into DbManager, previous step to a real DbManager

QuiqueOlaso:
Refactoring queries into DbManager, previous step to a real DbManager


andreasrosdalw:
Thanks. I will accept this improvement, for now.

However, for the future, I think Freeciv-web should use Hibernate:
https://github.com/freeciv/freeciv-web/issues/268

QuiqueOlaso:
Hmmmm, i also agree, but i was afraid: i though it would be rejected

andreasrosdalw:
If you want, please help this project by adding Hibernate support.

PR #282, opened 2019-07-26
Update CONTRIBUTING.md

andreasrosdalw:


cazfi:
I'd say progress should not be completely halted when nobody's doing the reviews. Maybe if there's no objections within 48h to the pull request, it's ok to merge.

zekoz:
The codebases are diverging quite a bit. I don't think there is much point in a common repo. I have made many radical changes that @andreasrosdal would never agree with, and the code quality of the fork is not up to any standard I'd accept.

andreasrosdalw:
@zekoz That is why we need separate GitHub repositories for each Freeciv-web instance. This freeciv-web repository needs to be generic Freeciv-web, while Freeciv.xyz needs to have a separate GitHub repository. 

For example, Freeciv.xyz could have github.com/freeciv/freeciv.xyz

zekoz:
Doesn't really address my point.

zekoz:
Did you get out of band approval to merge this?

andreasrosdalw:
@zekoz It's not merged yet, I updated based on the feedback given from cazfi. 

The general idea is to encourage multiple separate instances of Freeciv-web.
The freeciv/freeciv-web repo should be generic, and not used for one specific site.

andreasrosdalw:
> The codebases are diverging quite a bit. I don't think there is much point in a common repo.

The purpose of a generic common repo for Freeciv-web, is to have one single source to develop and improve Freeciv-web. Contributions to this common repo would benefit all Freeciv-web sites. Then specific Freeciv-web sites, such as freeciv.xyz could have their owns Git repos where their custom modifications are done.

cazfi (approved):

PR #303, opened 2020-02-18
Reenable CasperJS tests

lonemadmax:
Before merging, please consider this:
* The Travis build takes about twice as much time
* The tests are very general, they may only be useful to find if the client can start, not whether part of functionality work
* CasperJS is unmantained


andreasrosdalw:
Excellent 😊

andreasrosdalw:
Maybe we should start looking for a replacement for CasperJS?

andreasrosdalw:
https://github.com/freeciv/freeciv-web/issues/304

PR #309, opened 2020-05-19
CONTRIBUTING.md: Limit merge rights to maintainers only

cazfi:


andreasrosdalw (approved):

Lexxie9952:
Hey I know you did a lot of work to try to do this second capital, and now
you are getting all this ranting from me about it. It's a bad way to start
our new good relations for me to rant, and I'm sorry about it and
appreciate a lot the effort for this.

It is just that I know so many things about the consequences of this and I
was never asked. As we go forward I'm sure we will learn better how to work
together. I could have written more things about the importance and
consequences. Maybe from past experience you'd maybe post a line like "what
if we do like this ..." and see if someone has objection to it, maybe the
whole thing goes a lot better. I am just telling you though, there are 4 or
5 little things like this that if we un-diverge them were on the New
Superhighway of progress, and this is adding one more thing that puts us
off on the dirt road with weeds and rocks.

I know when I write something really big and abstract with ability to
handle all kinds of possibilities and put a lot of work into it, that I
will resist fiercely anyone telling me later to change it all. So I expect
you to resist this fiercely. Which makes me think the only effect of me
complaining on this will be to make relations worse for other things. How
can I convince you that there might be an easier way to handle this issue,
in the interests of so many other good things?

PR #312, opened 2020-07-14
More build system

kvilhaugsvik:


andreasrosdalw:
I'm not sure about this one. Can you add more information about why?

kvilhaugsvik:
> I'm not sure about this one. Can you add more information about why?

I'm assuming you are asking about the removal of apt-get upgrade: Installing upgrades can cause issues. Stuff we depend on that needs an upgrade will be upgraded as a part of installing our dependencies. The problem that caused me to remove it is that running an upgrade on Ubuntu focal on Travis sometimes will lock up the build.

kvilhaugsvik:
(Removal of the upgrade isn't enough to make building on Ubuntu focal work)

kvilhaugsvik:
Example: https://travis-ci.org/github/kvilhaugsvik/freeciv-web/builds/706530809

PR #314, opened 2020-07-30
Timeout flag conversion after 30 seconds.

kvilhaugsvik:
The conversion of the flags of Freeciv's bundled nations' from SVG can result in an eternal loop on some systems. Add a time out to stop it.


kvilhaugsvik:
The timeout is for each SVG file.

andreasrosdalw:
Excellent

PR #327, opened 2021-03-16
Update to Freeciv server freeciv/freeciv@2cd6d13c

cazfi:
Minor server update mainly to test the process.
Server's commit message dated Sun Jul 7 10:18:44 2019 +0300

Signed-off-by: Marko Lindqvist <cazfi74@gmail.com>

### DIFF
--- a/freeciv/version.txt
+++ b/freeciv/version.txt
@@ -1,9 +1,9 @@
 # The Git SHA hash for the commit to checkout from
 # https://github.com/freeciv/freeciv
 
-FCREV=3d94040eb30c36529bf7e6c4770136a78f8f75f3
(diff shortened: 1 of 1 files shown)


andreasrosdalw:
I recommend play-testing the game when upgrading server versions, since there could be breakages.

cazfi:
Yes. I stopped to that particular commit just because next one seems like one that requires work on web-client side.

kvilhaugsvik:
> Minor server update mainly to test the process

Just in case you already weren't aware of the check list: https://github.com/freeciv/freeciv-web/blob/develop/ADVANCED.md#upgrading-freeciv-webs-freeciv-version

PR #368, opened 2021-04-08
Fix lgtm alert

kvilhaugsvik:
Remove redundant check

### DIFF
--- a/freeciv-web/src/main/webapp/javascript/control.js
+++ b/freeciv-web/src/main/webapp/javascript/control.js
@@ -759,7 +759,7 @@ function ask_server_for_actions(punit)
 
   ptile = index_to_tile(punit['action_decision_tile']);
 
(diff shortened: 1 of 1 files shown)


andreasrosdalw:
Cool.

Will you help improve the 3D WebGL version? @kvilhaugsvik

kvilhaugsvik:
I'm still not in a position to work on the 3D web client. Even working on the 2D client is slow. At the moment a unit move in the 2d client takes 10 seconds in the web browser of the computer I develop on. I gave up trying to load the 3D page. (There is - as far as I know - no JavaScript JIT for the CPU in my main computer that is easy to install. I'm working on figuring out how to forward access to the virtual machine when Freeciv-web runs to a amd64 laptop.)

andreasrosdalw:
Du kan prøve å finne en bra brukt stasjonær PC på finn.

PR #422, opened 2021-10-30
Add freecivweb.org back to list of AGPL compliant sites

cazfi:
We've agreed that freecivweb.org is now, after the changes, compliant


kvilhaugsvik:
Approved. I think this means that you can merge it if you too view it as a minor change.

PR #447, opened 2022-03-22
Update various libraries

andreasrosdalw:
Update to jQuery UI v1.13.1. 
Update Vagrant to use image ubuntu jammy. 
Update various Java maven dependencies.


cazfi:
I need to backport one server fix for Jammy before switching to it.

andreasrosdalw:
Which backport is needed? @cazfi

cazfi:
osdn #44131 ... working on PR to contain it.

PR #456, opened 2022-04-16
Update to Freeciv server freeciv/freeciv@57665009ed

cazfi:


Lexxie9952:
I have no way to stop the madness, of your self-censorship from voluntarily non-knowingness, so I protest here. Since you atart to give opinions and things and are in an important role in the project, you need to consider the possibility of data collection on whether people are doing politics on you.

andreasrosdalw:
It's great that the server is updated in this way.

Lexxie, what are you talking about?

PR #465, opened 2022-05-08
3D Facelift by Lexxie

andreasrosdalw:
3D Facelift by Lexxie

### DIFF
--- a/freeciv-web/src/main/webapp/docs/controls.txt
+++ b/freeciv-web/src/main/webapp/docs/controls.txt
@@ -116,3 +116,8 @@ Remove forest
 
   Shift-Return: Turn done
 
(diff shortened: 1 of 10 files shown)


cazfi:
Can you make this just one commit? It doesn't make sense to intentionally introduce a bug to the version history by merging the first commit that brings in the bug, and the second one that fixes it, as separate commits.

andreasrosdal:
I would appreciate if you please merge this. Doing what you requested will require effort and energy which I would prefer to not prioritize on this (clean git commit history).

andreasrosdal:
I thought this was a nice improvement from Lexxie, that's all

cazfi:
I'm not sure that typing two commands to squash commits and to force push the branch again would have taken more effort than writing the comment that you won't do that, but ok. In similar situations in the future, do commit --amend in the first place - that actually saves you the effort of writing the extra commit message for an additional commit.

PR #480, opened 2022-06-23
Srvup

cazfi:


andreasrosdalw:
@cazfi You may submit these pull-requests to downstream [fciv-net](https://github.com/fciv-net/fciv-net) also, if you want. I sent you a invite. That way Fciv.net would get in sync with upstream, and you get to test your changes to freeciv-web on a real production server www.fciv.net.

PR #485, opened 2022-07-02
Prevent user from changing ruleset

cazfi:


andreasrosdalw:
I think the user would like to change ruleset

cazfi:
Do you have some better idea for a quick fix for the settings reset problem? Some of those publite settings are quite critical, and I don't have time for any extensive fix now.

cazfi:
> Do you have some better idea for a quick fix for the settings reset problem?

If not, I'm going to merge this.

PR #504, opened 2022-07-27
Srvup

cazfi:


cazfi:
The wrap setting that's part of this update suffers from a regression: https://osdn.net/projects/freeciv/ticket/45261

Will fold fix to that to relevant commit.

PR #512, opened 2022-08-17
Srvup

cazfi:


andreasrosdalw:
Is CMA possible in Freeciv-web now with this?

PR #551, opened 2022-11-05
Srvup

cazfi:


Lexxie9952:
This one is for Marko only, we are obviously doing nothing on
fractionalism strategy and I know pro-coders hate this kind of thing but
what ever happened to delegation leadership?  I watched upstream take a
clean hands politics-free approach during the zekoz-cheat-DMCA scandals
without knowing what's going on. I watch the forks happen from
overstretched resources preferring to code only and avoid human drama. I
watch stupid games like Forge of Empires get 1000x users from us. Who can I
talk to for making freeciv great?

cazfi:
Delegations? Well, two things make it hard. 1) There's nobody to delegate to (currently active members are myself, and Alina handling occasional patch). 2) Even if we had more active members, there's no supreme leader that could tell what people volunteer (army style) to do. Naturally I can always ask if someone could look at some ticket that might fall under their expertise, but that's it.

(I've been mostly away some days now myself, just noticing that work is piling up. For personal reasons my activity is likely low for some time)

Lexxie9952:
We want the same thing so, I will hear what you say first.

andreasrosdalw:
Please also don't forget the pretty new 3D version :)

PR #650, opened 2023-04-08
webclient: Start requesting generic "Clean" action

cazfi:


cazfi:
Need to have https://osdn.net/projects/freeciv/ticket/47839 backported, or the issue otherwise resolved, first.

PR #655, opened 2023-04-14
Sanitize username input

cazfi:


cazfi:
Warnings that CodeQL lists as "new" ones for the patch are not actually new. Those lines just have changed in a way that it doesn't recognize them, and lists those same warnings from the old version as "fixed" and as "new" from the new version.

PR #684, opened 2023-05-31
Correct extra removal cause values

cazfi:


andreasrosdalw:
What about this? https://github.com/freeciv/freeciv-web/blob/develop/freeciv-web/src/main/webapp/javascript/action_dialog.js#L878

cazfi:
Yes, that needs to be changed too. However, it's not breaking things just now, as ERM_CLEANPOLLUTION or ERM_CLEANFALLOUT is still (at this point of migration) always provided together with ERM_CLEAN.

PR #692, opened 2023-06-06
Make failure to install python modules non-fatal

cazfi:


andreasrosdalw:
I don't think this is a good idea, if the Python modules are used, they should be required.

cazfi:
What if we make python modules installation optional instead? So keep failure fatal if it fails in those cases where it's attempted, but not preventing builds when it's explicitly disabled.

PR #697, opened 2023-06-10
Lock nuclearwinter setting off

cazfi:


andreasrosdalw:
Where is the bottleneck? How can this be fixed? Does www.Fciv.net have the same problem?

cazfi:
I didn't investigate. Just had a freeciv-web freeze (or so I thought, it does recover eventually), and it turned out to be Nuclear Winter. You had already disabled Global Warming for similar reason.

PR #718, opened 2023-06-23
packhand.js: Declare map as global

cazfi:


andreasrosdalw:
map is already declared here:  https://github.com/freeciv/freeciv-web/blob/develop/freeciv-web/src/main/webapp/javascript/map.js#L21
So this looks like a mistake.

cazfi:
That declaration is not visible on packhand.js.

andreasrosdalw:
All the JavaScript files are bundled together by Closure compiler when the project is built using Maven. So there is no concept of variable scope or variable visibility in specific JavaScript files. Closure compiler will bundle and minify the JavaScript files into one file.

cazfi:
Doesn't help with the CodeQL warnings. Yes, it's only a warning, but I'd want to reduce amount of spam to make sure important ones won't go unnoticed among it.

andreasrosdalw:
CodeQL is incorrect in this case, it isn't aware of Closure compiler. Can you mark this case as false positive by CodeQL? I don't think it's a good idea to do this change to all global variables. Here the automated checks by CodeQL don't understand Closure compiler.

PR #723, opened 2023-06-29
Bug fix: deleting/loading saved files fails.

SiyuanQi:
Changes:
- Server saves the file in different possible formats, including '.xz' and '.zst'. Currently some logic hardcodes the filename to use the '.xz' extension, which causes failures when the file is saved as '.zst'. This happends when a game is loaded and saved again. See https://gist.github.com/SiyuanQi/f5890efc30c661b5212adc0fb233e8c1 for a screenshot.
- Keep the login logic consistent with LoginUser.java: https://github.com/freeciv/freeciv-web/blob/4752797b8ee7b3abef8e1a4dc431bde34b99e221/freeciv-web/src/main/java/org/freeciv/servlet/LoginUser.java#L83-L84

### DIFF
--- a/freeciv-web/src/main/java/org/freeciv/servlet/DeleteSaveGame.java
+++ b/freeciv-web/src/main/java/org/freeciv/servlet/DeleteSaveGame.java
@@ -32,9 +32,9 @@ import jakarta.servlet.*;
 import jakarta.servlet.http.*;
 import javax.sql.DataSource;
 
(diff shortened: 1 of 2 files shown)


cazfi:
> * Server saves the file in different possible formats, including '.xz' and '.zst'.

If you currently manage to save game in .zst format, that indicates another bug.

publite2/web.serv:
```
set compresstype xz
lock compresstype
```

